PR #49, opened 2021-03-24
create v3 migrator

NoahZinsmeister:


contracts/interfaces/ISelfPermit.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.7.5;
+
+/// @title Self Permit
+/// @notice Functionality to call permit on any EIP-2612-compliant token for use in the route
+interface ISelfPermit {
```
moodysalem, line 6: can we put this on master?

moodysalem (approved):
looks great, i guess any gas tests are kinda redundant but would be nice to have one just to know the overhead

contracts/V3Migrator.sol
```diff
@@ -0,0 +1,105 @@
...
+            LiquidityAmounts.getLiquidityForAmounts(
+                sqrtRatioX96,
+                TickMath.getSqrtRatioAtTick(params.tickLower),
+                TickMath.getSqrtRatioAtTick(params.tickUpper),
+                amount0V2,
+                amount1V2
+            );
+
+        // approve the position manager up to the maximum token amounts
+        IERC20(token0).approve(nonfungiblePositionManager, amount0V2);
```
moodysalem, line 74: oh `safeApprove`?
also this may fail if the approval is non-zero for some tokens (USDT)

contracts/V3Migrator.sol
```diff
@@ -0,0 +1,105 @@
...
+                    amount1Max: params.amount1Max,
+                    recipient: params.recipient,
+                    deadline: params.deadline
+                })
+            );
+
+        // if necessary, refund dust and clear allowance
+        if (amount0V3 < amount0V2) {
+            TransferHelper.safeTransfer(token0, msg.sender, amount0V2 - amount0V3);
+            IERC20(token0).approve(nonfungiblePositionManager, 0);
```
moodysalem, line 98: ah see this is how the approval is handled. that's probably the most elegant way to handle it

contracts/V3Migrator.sol
NoahZinsmeister: good point on `safeApprove`, i'm not even checking return values here, will add

Issue #54, opened 2021-03-29
allow specifying price limits in swaps


NoahZinsmeister:
- certainly need to be added to the quoter
- perhaps should be added to the router directly

considerations:
- to limit complexity, price limits would apply to the last/first swap of exact input/output trades, but this leads to a slightly strange API

PR #59, opened 2021-03-31
quote with price limit

moodysalem:
part of https://github.com/Uniswap/uniswap-v3-periphery/issues/54

### DIFF
--- a/contracts/lens/Quoter.sol
+++ b/contracts/lens/Quoter.sol
@@ -3,6 +3,7 @@ pragma solidity =0.7.6;
 pragma abicoder v2;
 
 import '@uniswap/v3-core/contracts/libraries/SafeCast.sol';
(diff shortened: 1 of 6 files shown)


contracts/lens/Quoter.sol
```diff
@@ -62,8 +58,24 @@ contract Quoter is IQuoter, IUniswapV3SwapCallback, PeripheryImmutableState, Per
...
+        }
+        return abi.decode(reason, (uint256));
+    }
+
+    /// @dev Quotes a single exact input swap
+    function quoteExactInputSingle(
+        uint256 amountIn,
+        bytes memory path,
+        uint160 sqrtPriceLimitX96
+    ) public returns (uint256 amountOut) {
```
moodysalem, line 78: feels like we should return both since there is a price limit

PR #60, opened 2021-03-31
NFT metadata: allow token0 numerator for certain pair ratios

ewilz:


contracts/NonfungibleTokenPositionDescriptor.sol
```diff
@@ -11,7 +12,21 @@ import './libraries/NFTDescriptor.sol';
...
 /// @title Describes NFT token positions
 /// @notice Produces a string containing the data URI for a JSON metadata string
-contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescriptor {
+contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescriptor, Ownable {
+    // tokens that take priority as ratio numerator (such as stablecoins)
+    mapping(address => bool) public ratioNumeratorTokens;
+    // tokens that take priorty as ratio denominator (such as WETH)
+    mapping(address => bool) public ratioDenominatorTokens;
+
+    constructor(address[] memory numeratorTokens, address[] memory denominatorTokens) Ownable() {
```
moodysalem, line 21: this state could complicate the upgradeability

any way to do this with immutability, i.e. so that it's in the code? tbh i don't know of one but if there's another strategy would be great

contracts/NonfungibleTokenPositionDescriptor.sol
```diff
@@ -11,7 +12,21 @@ import './libraries/NFTDescriptor.sol';
 
 /// @title Describes NFT token positions
 /// @notice Produces a string containing the data URI for a JSON metadata string
-contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescriptor {
+contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescriptor, Ownable {
```
moodysalem, line 15: idt this should be ownable

contracts/NonfungibleTokenPositionDescriptor.sol
```diff
@@ -48,4 +64,24 @@ contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescript
...
+        } else if (ratioNumeratorTokens[token0]) {
+            return true;
+        } else if (ratioDenominatorTokens[token1]) {
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    function addRatioNumeratorToken(address token) external onlyOwner() {
```
ewilz, line 80: aww crap....just realizing that if we use the zeppelin `TransparentProxy`  governance cannot both be the admin for upgradeability and _also_ call to the implementation contract... we may have to overwrite that if we want governance to be able to assign more of these..

contracts/NonfungibleTokenPositionDescriptor.sol
ewilz: I guess I imagined it might be something where we'd want to add more as we saw fit. Though having governance call the implementation is a tiny bit tricky per this comment (https://github.com/Uniswap/uniswap-v3-periphery/pull/60/files#r605287205)

why would this state complicated upgradeability?

contracts/NonfungibleTokenPositionDescriptor.sol
ewilz: If we want governance to be able to add more tokens, some sort of ownable is necessary, but if we don't want to add more tokens in the future, then we can take this out

contracts/NonfungibleTokenPositionDescriptor.sol
```diff
@@ -11,7 +13,21 @@ import './libraries/NFTDescriptor.sol';
...
+        address token;
+        int256 priority;
+    }
+
+    // tokens that take priority order in price ratio - higher integers get numerator prioriy
+    mapping(address => int256) public tokenRatioPriority;
+
+    function initialize(TokenRatioOrderPriority[] memory tokens) public initializer() {
+        for (uint256 i = 0; i < tokens.length; i++) {
+            tokenRatioPriority[tokens[i].token] = tokens[i].priority;
```
ewilz, line 27: maybe should trigger an event for each token to have a record of which tokens were recorded...(helpful for making changes in the future)

moodysalem (approved):
lgtm, just gotta find out why the gas snapshots changed

test/__snapshots__/SwapRouter.gas.spec.ts.snap
```diff
@@ -1,19 +1,19 @@
...
 
 exports[`SwapRouter gas tests #exactInput 0 -> 1 1`] = `111157`;
 
 exports[`SwapRouter gas tests #exactInput 0 -> 1 minimal 1`] = `104691`;
 
-exports[`SwapRouter gas tests #exactInput 0 -> WETH9 1`] = `137996`;
+exports[`SwapRouter gas tests #exactInput 0 -> WETH9 1`] = `138141`;
 
-exports[`SwapRouter gas tests #exactInput WETH9 -> 0 1`] = `111324`;
+exports[`SwapRouter gas tests #exactInput WETH9 -> 0 1`] = `111179`;
```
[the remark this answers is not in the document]
ewilz: mm..not sure..this file changed a bit in my last PR too. Happened on merge

PR #68, opened 2021-04-01
migrator changes

NoahZinsmeister:
add ability to refund in eth

change slippage to v3 liquidity min

add refund tests

### DIFF
--- a/contracts/V3Migrator.sol
+++ b/contracts/V3Migrator.sol
@@ -22,9 +22,11 @@ import './interfaces/external/IWETH9.sol';
 contract V3Migrator is IV3Migrator, PeripheryImmutableState, Multicall, SelfPermit {
     address public immutable nonfungiblePositionManager;
 
-    constructor(address _factory, address _WETH9, address _nonfungiblePositionManager)
-        PeripheryImmutableState(_factory, _WETH9)
-    {
+    constructor(
+        address _factory,
+        address _WETH9,
+        address _nonfungiblePositionManager
+    ) PeripheryImmutableState(_factory, _WETH9) {
         nonfungiblePositionManager = _nonfungiblePositionManager;
     }
 
--- a/contracts/V3Migrator.sol
+++ b/contracts/V3Migrator.sol
@@ -13,17 +13,23 @@ import './libraries/TransferHelper.sol';
 
 import './libraries/LiquidityAmounts.sol';
 import './interfaces/IV3Migrator.sol';
+import './base/PeripheryImmutableState.sol';
 import './base/Multicall.sol';
 import './base/SelfPermit.sol';
+import './interfaces/external/IWETH9.sol';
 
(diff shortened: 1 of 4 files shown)


contracts/V3Migrator.sol
```diff
@@ -13,19 +13,27 @@ import './libraries/TransferHelper.sol';
...
-        factory = _factory;
+    constructor(
+        address _factory,
+        address _WETH9,
+        address _nonfungiblePositionManager
+    ) PeripheryImmutableState(_factory, _WETH9) {
         nonfungiblePositionManager = _nonfungiblePositionManager;
     }
 
+    receive() external payable {
```
moodysalem, line 33: should this be inherited?

contracts/V3Migrator.sol
```diff
@@ -43,7 +51,7 @@ contract V3Migrator is IV3Migrator, Multicall, SelfPermit {
 
     function migrate(MigrateParams calldata params) external override {
         // burn v2 liquidity to this address
-        IUniswapV2Pair(params.pair).transferFrom(msg.sender, params.pair, params.liquidityV2);
+        IUniswapV2Pair(params.pair).transferFrom(msg.sender, params.pair, params.liquidityToMigrate);
         (uint256 amount0V2, uint256 amount1V2) = IUniswapV2Pair(params.pair).burn(address(this));
 
         // get underlying tokens
```
moodysalem, line 57: probably cheaper to take v2Factory, token0, token1 and compute the pair, than make these two calls (although idk if the bytecode hash is same for all forks, probably not. should we prioritize uniswap gas cost tho?)

contracts/V3Migrator.sol
```diff
@@ -85,21 +93,35 @@ contract V3Migrator is IV3Migrator, Multicall, SelfPermit {
...
                     recipient: params.recipient,
                     deadline: params.deadline
                 })
             );
 
-        // if necessary, refund dust and clear allowance
+        // if necessary, clear allowance and refund dust
         if (amount0V3 < amount0V2) {
-            TransferHelper.safeTransfer(token0, msg.sender, amount0V2 - amount0V3);
             IERC20(token0).approve(nonfungiblePositionManager, 0);
```
moodysalem, line 105: safeApprove?

moodysalem (approved):
sc code lgtm

contracts/V3Migrator.sol
NoahZinsmeister: i don't want to add the sweep methods into here though, as we have an explicit param for handling eth and i don't want to confuse people

contracts/V3Migrator.sol
NoahZinsmeister: for interop i prefer explicit pair address. perhaps we could just pass the token addresses in calldata?

moodysalem: yeah that works

Issue #56, opened 2021-03-31
NFT Metadata improvements 


ewilz:
- [x] Display price/tokens in priority order for ETH & stablecoins (token/ETH, stablecoin/token) https://github.com/Uniswap/uniswap-v3-periphery/pull/60
- [x] Submit final draft for description text (https://github.com/Uniswap/uniswap-v3-periphery/pull/94)
- [x] Fuzz Test (https://github.com/Uniswap/uniswap-v3-periphery/pull/93)
- [ ] Account for token symbol with quotes 0_o would break JSON(https://github.com/Uniswap/uniswap-v3-periphery/pull/104)

ewilz:
https://github.com/Uniswap/uniswap-v3-periphery/pull/93

PR #70, opened 2021-04-04
NFTDescriptor token priority data in bytecode

ewilz:
https://github.com/Uniswap/uniswap-v3-periphery/issues/56


contracts/base/ChainId.sol
```diff
@@ -0,0 +1,16 @@
...
+
+/// @title Function for getting the chainid
+/// @dev Base contract that is overridden for tests
+abstract contract ChainId {
+    /// @dev Method that exists purely to be overridden for tests
+    /// @return The chainId
+    function _chainid() internal view virtual returns (uint256) {
+        uint256 chainID;
+        assembly {
+            chainID := chainid()
```
moodysalem, line 12: if you make this a one liner it will be inlined into the calling fn

also naming can just be get() and it should be a library

contracts/NonfungibleTokenPositionDescriptor.sol
```diff
@@ -67,11 +59,33 @@ contract NonfungibleTokenPositionDescriptor is INonfungibleTokenPositionDescript
...
-        tokenRatioPriority[token.token] = token.priority;
-        emit UpdateTokenRatioPriority(token.token, token.priority);
+    function tokenRatioPriority(address token) public view returns (int256) {
+        if (token == WETH9) {
+            return -100;
+        }
+        if (_chainid() == 1) {
+            if (token == 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48) {
+                // USDC
+                return 250;
```
moodysalem, line 72: put these addresses in constants like in the test

also you can wrap the values in an enum for readability

moodysalem (approved):
small comments

PR #72, opened 2021-04-05
Test validationCallback reverts when called by address other than pair

ewilz:
V3 Pool

### DIFF
--- a/contracts/libraries/CallbackValidation.sol
+++ b/contracts/libraries/CallbackValidation.sol
@@ -31,6 +31,6 @@ library CallbackValidation {
         returns (IUniswapV3Pool pool)
     {
         pool = IUniswapV3Pool(PoolAddress.computeAddress(factory, poolKey));
-        require(msg.sender == address(pool));
+        require(msg.sender == address(pool), 'Callback not called from pool');
     }
 }
--- a/contracts/libraries/CallbackValidation.sol
+++ b/contracts/libraries/CallbackValidation.sol
@@ -31,6 +31,6 @@ library CallbackValidation {
         returns (IUniswapV3Pool pool)
     {
         pool = IUniswapV3Pool(PoolAddress.computeAddress(factory, poolKey));
-        require(msg.sender == address(pool), 'Callback not called from pool');
+        require(msg.sender == address(pool));
     }
 }
--- a/contracts/test/TestCallbackValidation.sol
+++ b/contracts/test/TestCallbackValidation.sol
@@ -3,7 +3,7 @@ pragma solidity =0.7.6;
 
 import '../libraries/CallbackValidation.sol';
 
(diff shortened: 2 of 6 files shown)


contracts/test/TestCallbackValidation.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.7.6;
+
+import '../libraries/CallbackValidation.sol';
+
+contract TestCallbackValdidation {
```
NoahZinsmeister, line 6: nit: typo

contracts/libraries/CallbackValidation.sol
```diff
@@ -31,6 +31,6 @@ library CallbackValidation {
         returns (IUniswapV3Pool pool)
     {
         pool = IUniswapV3Pool(PoolAddress.computeAddress(factory, poolKey));
-        require(msg.sender == address(pool));
+        require(msg.sender == address(pool), 'Callback not called from pool');
```
NoahZinsmeister, line 34: our general MO around error messages in core was to remove them if they're not expected to be hit in a normal case, and since this is only to protect against malicious behavior, perhaps not worth the gas? makes the new test that much less obvious, unfortunately

NoahZinsmeister (approved):

contracts/test/TestCallbackValidation.sol
ewilz: 😂 that's a pretty bad one, right in the contract name

contracts/libraries/CallbackValidation.sol
ewilz: this makes sense

test/__snapshots__/SwapRouter.gas.spec.ts.snap
```diff
@@ -6,26 +6,26 @@ exports[`SwapRouter gas tests #exactInput 0 -> 1 1`] = `111223`;
...
+exports[`SwapRouter gas tests #exactInputSingle 0 -> WETH9 1`] = `137616`;
 
 exports[`SwapRouter gas tests #exactInputSingle WETH9 -> 0 1`] = `110878`;
 
 exports[`SwapRouter gas tests #exactOutput 0 -> 1 -> 2 1`] = `177437`;
 
 exports[`SwapRouter gas tests #exactOutput 0 -> 1 1`] = `115743`;
 
-exports[`SwapRouter gas tests #exactOutput 0 -> WETH9 1`] = `139336`;
+exports[`SwapRouter gas tests #exactOutput 0 -> WETH9 1`] = `139324`;
```
ewilz, line 23: maybe the solution to cutting gas costs it writing a ton of random test contracts that magically bring down gas metrics

Issue #73, opened 2021-04-05
current mint params make it difficult to mint the exact amount of desired liquidity


moodysalem:
https://github.com/Uniswap/uniswap-v3-periphery/blob/33825d2570eb939ea44abc728e554634368dcd42/contracts/interfaces/INonfungiblePositionManager.sol#L46-L69

compared to the v2 router interface, which takes the desired amounts and min amounts then computes the liquidity amount to mint, this is interface is not as nice because it makes it difficult to spend the desired amount in the case of 0 slippage.

PR #75, opened 2021-04-05
fix minting interface to take token amounts and compute liquidity

moodysalem:
fixes https://github.com/Uniswap/uniswap-v3-periphery/issues/73

### DIFF
--- a/contracts/base/LiquidityManagement.sol
+++ b/contracts/base/LiquidityManagement.sol
@@ -19,8 +19,6 @@ abstract contract LiquidityManagement is IUniswapV3MintCallback, PeripheryImmuta
     struct MintCallbackData {
         PoolAddress.PoolKey poolKey;
         address payer;
-        uint256 amount0Min;
-        uint256 amount1Min;
     }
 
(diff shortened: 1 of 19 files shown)


contracts/base/LiquidityManagement.sol
```diff
@@ -58,15 +67,34 @@ abstract contract LiquidityManagement is IUniswapV3MintCallback, PeripheryImmuta
...
             params.tickLower,
             params.tickUpper,
-            params.amount,
-            abi.encode(MintCallbackData({poolKey: poolKey, payer: msg.sender}))
+            liquidity,
+            abi.encode(
+                MintCallbackData({
+                    poolKey: poolKey,
+                    payer: msg.sender,
+                    amount0Min: params.amount0Min,
```
moodysalem, line 94: thinking is it's slightly cheaper in case of revert to do it in the callback

NoahZinsmeister: i think we should optimize for the non-reverting case

NoahZinsmeister (approved):

PR #80, opened 2021-04-06
SVG Image first iteration (circles / token-based colors)

ewilz:


moodysalem:
consider base64 encoding it so you don't have to worry about escaping quotes (you will need to write the solidity for that, fun)

test/Base64.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { ethers } from 'hardhat'
 import { expect } from './shared/expect'
+import { base64Encode } from './shared/base64Encode'
```
ewilz, line 3: ^--- well....it's there

moodysalem (approved):

PR #81, opened 2021-04-06
make migrate slippage params on amounts

NoahZinsmeister:


contracts/V3Migrator.sol
```diff
@@ -6,6 +6,7 @@ import '@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol';
 import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
 import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
 import './interfaces/INonfungiblePositionManager.sol';
+import '@openzeppelin/contracts/math/SafeMath.sol';
```
moodysalem, line 9: can use lowgassafemath from core

contracts/V3Migrator.sol
```diff
@@ -49,37 +52,44 @@ contract V3Migrator is IV3Migrator, PeripheryImmutableState, Multicall, SelfPerm
...
     function migrate(MigrateParams calldata params) external override {
+        require(params.percentageToMigrate > 0, 'Percentage too small');
+        require(params.percentageToMigrate <= 100, 'Percentage too large');
+
         // burn v2 liquidity to this address
         IUniswapV2Pair(params.pair).transferFrom(msg.sender, params.pair, params.liquidityToMigrate);
         (uint256 amount0V2, uint256 amount1V2) = IUniswapV2Pair(params.pair).burn(address(this));
 
+        // calculate the amounts to migrate to v3
+        uint256 amount0V2ToMigrate = amount0V2.mul(params.percentageToMigrate).div(100);
```
moodysalem, line 63: .div(100) can just be `/ 100`

contracts/V3Migrator.sol
```diff
@@ -3,14 +3,11 @@ pragma solidity =0.7.6;
...
 import '@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol';
-import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
-import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
 import './interfaces/INonfungiblePositionManager.sol';
 
-import './libraries/PoolAddress.sol';
 import './libraries/TransferHelper.sol';
 
-import './libraries/LiquidityAmounts.sol';
+import '@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol';
```
moodysalem, line 10: nit: move up with others?

contracts/V3Migrator.sol
```diff
@@ -49,37 +48,52 @@ contract V3Migrator is IV3Migrator, PeripheryImmutableState, Multicall, SelfPerm
...
+                    amount0Min: params.amount0Min,
+                    amount1Min: params.amount1Min,
                     recipient: params.recipient,
                     deadline: params.deadline
                 })
             );
 
-        require(liquidityV3 >= params.liquidityV3Min, 'Excessive price impact');
+        // special slippage checks for migrating out of range
+        if (params.amount0Min == 0) {
```
moodysalem, line 85: boooo special values. can we just take amount0Desired, amount1Desired and that serves as slippage checks for the v1 burn/v2 mint?

moodysalem: (also would replace the percentage thing)

NoahZinsmeister: i don't know...

ignoring the percentage stuff for now - are you imagining passing the desired amounts as (1-slippage) * current v2 value (adjusted for your v3 position)? because the problem with that is that you're not adding the maximum amount of liquidity unless the v2/v3 prices change s.t. you're exactly at one of your slippage limits.

if the desired amounts are just the current v2 value (adjusted for your v3 position), and price movement will cause your migrate to fail

NoahZinsmeister: and if you're doing the percentage thing, then the desired amounts _really_ don't protect you unless you also pass min amounts, because you're fully exposed to price changes up until your percentage

NoahZinsmeister: i don't think these checks are crazy at all tbh, they're simply there to enforce that an out-of-range position at submit time is still out-of-range at runtime, if you want that

moodysalem: true. since the user is in both assets to begin with, they might prefer the migration to continue than for it to fail because the prices moved slightly into the tick range. maybe just remove these checks since they change the meaning of the parameter and are maybe not useful.

moodysalem (approved):

PR #83, opened 2021-04-08
add exact output check

NoahZinsmeister:


contracts/SwapRouter.sol
```diff
@@ -189,23 +196,20 @@ contract SwapRouter is
...
+            params.amountOut,
+            params.recipient,
+            params.sqrtPriceLimitX96,
+            SwapCallbackData({
+                path: abi.encodePacked(params.tokenOut, params.fee, params.tokenIn),
+                payer: msg.sender
+            })
+        );
 
+        require(amountIn <= params.amountInMaximum, 'Too much requested');
```
NoahZinsmeister, line 210: this saves gas in failure case

contracts/SwapRouter.sol
```diff
@@ -179,6 +179,13 @@ contract SwapRouter is
                     : sqrtPriceLimitX96,
                 abi.encode(data)
             );
+
+        uint256 amountOutReceived;
+        (amountIn, amountOutReceived) = zeroForOne
+            ? (uint256(amount0Delta), uint256(-amount1Delta))
+            : (uint256(amount1Delta), uint256(-amount0Delta));
+        // it's technically possible to not receive the full output amount
+        require(amountOutReceived == amountOut);
```
moodysalem, line 188: hmm actually this breaks exactOutputSingle with meaningful price limits but i guess that's ok

NoahZinsmeister: oh i didn't even think of that to be honest...odd, but i guess they can just use exactInputSingle? is this an argument to remove the price limit arg from exactOutputSingle?

NoahZinsmeister: addressed by only running the check when price limit isn't specified

PR #84, opened 2021-04-08
IQuoter, IPeripheryPayments, IPeripheryImmutableState natspec

snreynolds:


contracts/interfaces/IPeripheryImmutableState.sol
```diff
@@ -4,9 +4,9 @@ pragma solidity >=0.5.0;
 /// @title Immutable state
 /// @notice Functions that return immutable state of the router
 interface IPeripheryImmutableState {
-    /// @notice Returns the address of the Uniswap V3 factory
+    /// @return Returns the address of the Uniswap V3 factory
     function factory() external view returns (address);
 
-    /// @notice Returns the address of WETH9
+    /// @return Returns the address of WETH9
```
moodysalem, line 10: kindof odd use of `@return`

PR #89, opened 2021-04-14
fix exact input payer bug

### LINKED ISSUE #87: payer in multihop should be router except for first swap
https://github.com/Uniswap/uniswap-v3-periphery/blob/8ad273f45cd42e04b5f8f15b6f4da57827122273/contracts/SwapRouter.sol#L145

i guess tests are passing because all tokens are approved

NoahZinsmeister:
closes #87

### DIFF
--- a/contracts/SwapRouter.sol
+++ b/contracts/SwapRouter.sol
@@ -144,7 +144,7 @@ contract SwapRouter is
                 0,
                 SwapCallbackData({
                     path: params.path.getFirstPool(), // only the first pool in the path is necessary
-                    payer: payer 
+                    payer: payer
                 })
             );
 
--- a/contracts/SwapRouter.sol
+++ b/contracts/SwapRouter.sol
@@ -148,10 +148,9 @@ contract SwapRouter is
                 })
             );
 
-            if (payer != address(this)) payer = address(this); // this contract pays for subsequent hops
-
             // decide whether to continue or terminate
             if (hasMultiplePools) {
+                payer = address(this); // at this point, the caller has paid
                 params.path = params.path.skipToken();
             } else {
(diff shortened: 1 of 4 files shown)


moodysalem (approved):

contracts/SwapRouter.sol
```diff
@@ -142,10 +144,12 @@ contract SwapRouter is
                 0,
                 SwapCallbackData({
                     path: params.path.getFirstPool(), // only the first pool in the path is necessary
-                    payer: msg.sender
+                    payer: payer
                 })
             );
 
+            if (payer != address(this)) payer = address(this); // this contract pays for subsequent hops
```
moodysalem, line 151: wait this is an odd way to do it, it has a branch every time plus an mstore, we could just use a ternary in the payer?

moodysalem: if you just remove the if branch i'd guess it's cheaper and still correct

NoahZinsmeister: well you need something like this in any case. pseudocode:

hasPaid = false
// ...
payer: hasPaid ? address(this) : msg.sender
// ...
if (!hasPaid) hasPaid = false // if statement is so we only mstore once

so this is 2 MSTOREs, and 2 conditionals / loop iteration

my version is 2 MSTOREs, and 1 conditional / loop iteration

does that seem right?

moodysalem: i suspect the if statement creates a jump which is more expensive than just setting it every time

NoahZinsmeister: ok, will check the gas costs and merge whichever's cheaper

PR #93, opened 2021-04-14
NFT Descriptor: decimal strings fuzz test

ewilz:


test/NFTDescriptor.spec.ts
```diff
@@ -389,6 +396,46 @@ describe('NFTDescriptor', () => {
           expect(await nftDescriptor.fixedPointToDecimalString(ratio, 24, 5)).to.eq('1.0000')
         })
       })
+
+      it('some fuzz', async () => {
+        const random3To21 = () => {
+          return Math.floor(3 + ((Math.random() * 100) % 18))
+        }
+        const random1To20 = () => {
```
moodysalem, line 404: looks like you probably just want something like `rand(min: number, maxExclusive: number)`

test/NFTDescriptor.spec.ts
```diff
@@ -389,6 +396,46 @@ describe('NFTDescriptor', () => {
...
+              .mul(TEN.pow(decimalDiff))
+              .lt(HIGHEST_SQRT_RATIO)
+          ) {
+            inputs.push([BigNumber.from(`0x${value.toString('hex')}`), decimals0, decimals1])
+            i++
+          }
+        }
+
+        for (let i in inputs) {
+          let input: any
```
moodysalem, line 431: why any?

test/shared/formatSqrtRatioX96.ts
```diff
@@ -1,21 +1,33 @@
...
 ): string {
-  Decimal.set({ precision: 5, toExpPos: 9_999_999, toExpNeg: -9_999_999, rounding: Decimal.ROUND_HALF_CEIL })
+  Decimal.set({ precision: 5, toExpPos: 9_999_999, toExpNeg: -9_999_999, rounding: Decimal.ROUND_HALF_EVEN })
 
-  const decRatio = new Decimal(sqrtRatioX96.toString())
-
-  let ratio = decRatio.div(new Decimal(2).pow(96)).pow(2)
+  let ratio
+  let ratioNum
+  let ratioBN = sqrtRatioX96.mul(sqrtRatioX96).div(TWO.pow(64))
```
moodysalem, line 17: ideally you just use another mostly self contained implementation that we should expect to be correct, rather than reimplement the solidity logic in js, but as long as we know this implementation is correct it's nbd. 

i think the reason it was off before could be that precision is set to 5 on line 12, but i think i did that for formatting the text. it might be possible to get Decimal to be accurate with higher precision though

ewilz: Even without precision, the Decimals library calculations are not as accurate as BigNumber or even JS number and will have intermittent errors with the full number being slightly off.

The way it is currently now, there are 0 discrepancies, except when doing token decimals adjustments which once again, do very large multiplications and divisions which impact the final result by a bit. The only way to get the same exact result is to do those same mul/divs in javascript. Which honestly kind of sucks..... 

I think we know the implementation is correct now, but it WASN'T. And I knew that thanks to not copying every little solidity thing...

PR #94, opened 2021-04-16
Finalize NFT description metadata

ewilz:
This NFT represents a liquidity position in a Uniswap V3 TEST-TEST pool. The owner of this NFT can modify or redeem the position.

Pool Address: 0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
TEST Address: 0xe7f1725e7734ce288f8367e1bb143e90bb3f0512
TEST Address: 0x9fe46736679d2d9a65f0992f2272de9f3c7fa6e0
Fee Tier: 0.3%
Token ID: 123

⚠️ DISCLAIMER: Due diligence is imperative when assessing this NFT. Make sure token addresses match the expected tokens, as token symbols may be imitated."

### DIFF
--- a/contracts/libraries/NFTDescriptor.sol
+++ b/contracts/libraries/NFTDescriptor.sol
@@ -24,7 +24,7 @@ library NFTDescriptor {
 
     struct ConstructTokenURIParams {
         uint256 tokenId;
-        address quoteTokenAddr;
+        address quoteTokenAddress;
         address baseTokenAddr;
         string quoteTokenSymbol;
         string baseTokenSymbol;
@@ -82,7 +82,7 @@ library NFTDescriptor {
                     '\\n',
                     params.quoteTokenSymbol,
                     ' Address: ',
-                    addressToString(params.quoteTokenAddr),
+                    addressToString(params.quoteTokenAddress),
                     '\\n',
                     params.baseTokenSymbol,
                     ' Address: ',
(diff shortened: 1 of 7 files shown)


contracts/libraries/NFTDescriptor.sol
```diff
@@ -70,15 +73,28 @@ library NFTDescriptor {
         string memory description =
             string(
                 abi.encodePacked(
-                    'Represents a liquidity position in a Uniswap V3 pool. Redeemable for owed reserve tokens.',
-                    '\\nliquidity: ',
-                    uint256(params.liquidity).toString(),
-                    '\\npoolAddress: ',
+                    'This NFT represents a liquidity position in a Uniswap V3 ',
+                    token1Symbol,
```
moodysalem, line 77: do we need to do ordering of these symbols the same way we do prices? otherwise the prices may be confused to be flipped

moodysalem: e.g. we show ETH/DAI and then prices 1500-2500

ewilz: I've attempted to order them consistently throughout. Am I missing something?
(see line 44)

moodysalem: ah gotcha, yeah better to name them quoteToken and baseToken or something like that

moodysalem (approved):

Issue #22, opened 2021-03-09
Oracle library


moodysalem:
Need a library that returns TWAP for base currency/quote currency over some time period as a `FixedPoint128` number

Considerations
- how to choose between multiple pools
- other kinds of averages (EMA)

leonardgee:
This will be quite useful, is there any update on this? Also where can we find the oracle implementation for Uniswap v3?

alejoamiras:
Right now I think, althought the Oracle Library MVP is pretty cool, it's not procution / protocol integration ready. I'm trying to implement a programatical way of selecting the best pool, why not getting `!OLD` or being open to an attack, and couldn't find an easy anser (and don't have time to learn all the uni v3 codebase). I know in #96 @NoahZinsmeister talked about "aggregating over multiple fee tiers", is there an ETA on this ? 😭 
Great work, nonetheless.

moodysalem:
We now have https://github.com/Uniswap/uniswap-v3-periphery/blob/b771ff9a20a0fd7c3233df0eb70d4fa084766cde/contracts/libraries/OracleLibrary.sol

I agree we could build something opinionated on how to weight ticks across fee levels

PR #96, opened 2021-04-16
Implement oracle library

smsunarto:
#22

### DIFF
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity =0.7.6;
+
+import '@uniswap/v3-core/contracts/libraries/FullMath.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@uniswap/v3-core/contracts/interfaces/pool/IUniswapV3PoolDerivedState.sol';
+import '@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol';
+import 'hardhat/console.sol';
+
+library OracleLibrary {
+    function consult(
+        uint32 time,
+        address poolAddress,
+        uint32 startBlockTimestamp,
+        uint32 endBlockTimestamp
+    ) public view returns (uint256 ratioX128) {
+        require(startBlockTimestamp < endBlockTimestamp, 'Bad range');
+        IUniswapV3PoolDerivedState oracle = IUniswapV3PoolDerivedState(poolAddress);
+
+        uint32[] memory secondAgos = new uint32[](2);
+        secondAgos[0] = time - startBlockTimestamp;
+        secondAgos[1] = time - endBlockTimestamp;
+
+        (int56[] memory tickCumulatives, ) = oracle.observe(secondAgos);
+        int24 tick = int24((tickCumulatives[1] - tickCumulatives[0]) / int56(endBlockTimestamp - startBlockTimestamp));
+
+        uint160 sqrtRatioX96 = TickMath.getSqrtRatioAtTick(tick);
+        ratioX128 = FullMath.mulDiv(sqrtRatioX96, sqrtRatioX96, 1 << 64);
+    }
+}
--- a/contracts/test/OracleTest.sol
+++ b/contracts/test/OracleTest.sol
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.7.6;
+pragma abicoder v2;
+
+import '@uniswap/v3-core/contracts/libraries/Oracle.sol';
+import '../libraries/OracleLibrary.sol';
+
+contract OracleTest {
+    using Oracle for Oracle.Observation[65535];
+
+    Oracle.Observation[65535] public observations;
+
+    uint32 public time;
+    int24 public tick;
+    uint128 public liquidity;
+    uint16 public index;
+    uint16 public cardinality;
+    uint16 public cardinalityNext;
+
+    struct InitializeParams {
+        uint32 time;
+        int24 tick;
+        uint128 liquidity;
+    }
+
+    function initialize(InitializeParams calldata params) external {
+        require(cardinality == 0, 'already initialized');
+        time = params.time;
+        tick = params.tick;
+        liquidity = params.liquidity;
+        (cardinality, cardinalityNext) = observations.initialize(params.time);
+    }
+
+    function advanceTime(uint32 by) public {
+        time += by;
+    }
+
+    struct UpdateParams {
+        uint32 advanceTimeBy;
+        int24 tick;
+        uint128 liquidity;
+    }
+
+    // write an observation, then change tick and liquidity
+    function update(UpdateParams calldata params) external {
+        advanceTime(params.advanceTimeBy);
+        (index, cardinality) = observations.write(index, time, tick, liquidity, cardinality, cardinalityNext);
+        tick = params.tick;
+        liquidity = params.liquidity;
+    }
+
+    function batchUpdate(UpdateParams[] calldata params) external {
+        // sload everything
+        int24 _tick = tick;
+        uint128 _liquidity = liquidity;
+        uint16 _index = index;
+        uint16 _cardinality = cardinality;
+        uint16 _cardinalityNext = cardinalityNext;
+        uint32 _time = time;
+
+        for (uint256 i = 0; i < params.length; i++) {
+            _time += params[i].advanceTimeBy;
+            (_index, _cardinality) = observations.write(
+                _index,
+                _time,
+                _tick,
+                _liquidity,
+                _cardinality,
+                _cardinalityNext
+            );
+            _tick = params[i].tick;
+            _liquidity = params[i].liquidity;
+        }
+
+        // sstore everything
+        tick = _tick;
+        liquidity = _liquidity;
+        index = _index;
+        cardinality = _cardinality;
+        time = _time;
+    }
+
+    function grow(uint16 _cardinalityNext) external {
+        cardinalityNext = observations.grow(cardinalityNext, _cardinalityNext);
+    }
+
+    function observe(uint32[] calldata secondsAgos)
+        external
+        view
+        returns (int56[] memory tickCumulatives, uint160[] memory liquidityCumulatives)
+    {
+        return observations.observe(time, secondsAgos, tick, index, liquidity, cardinality);
+    }
+
+    function getGasCostOfObserve(uint32[] calldata secondsAgos) external view returns (uint256) {
+        (uint32 _time, int24 _tick, uint128 _liquidity, uint16 _index) = (time, tick, liquidity, index);
+        uint256 gasBefore = gasleft();
+        observations.observe(_time, secondsAgos, _tick, _index, _liquidity, cardinality);
+        return gasBefore - gasleft();
+    }
+
+    function consult(
+        address poolAddress,
+        uint32 startBlockTimestamp,
+        uint32 endBlockTimestamp
+    ) public view returns (uint256 ratioX128) {
+        ratioX128 = OracleLibrary.consult(time, poolAddress, startBlockTimestamp, endBlockTimestamp);
+    }
+}
--- a/test/OracleLibrary.spec.ts
+++ b/test/OracleLibrary.spec.ts
@@ -1,0 +1,127 @@
+import { expect } from 'chai'
+import { OracleTest } from '../typechain/OracleTest'
+import { TickMathTest } from '../typechain/TickMathTest'
+import { ethers, waffle } from 'hardhat'
+import formatSqrtRatioX96 from './shared/formatSqrtRatioX96'
+import { BigNumber } from 'ethers'
+
+describe('OracleLibrary', () => {
+  const [wallet, other] = waffle.provider.getWallets()
+
+  let loadFixture: ReturnType<typeof waffle.createFixtureLoader>
+  let oracle: OracleTest
+  let tickMath: TickMathTest
+
+  before('create fixture loader', async () => {
+    loadFixture = waffle.createFixtureLoader([wallet, other])
+  })
+
+  beforeEach('deploy test oracle and tick math', async () => {
+    const fixtures = await loadFixture(oracleFixture)
+    oracle = fixtures['oracle']
+    tickMath = fixtures['tickMath']
+  })
+
+  const oracleFixture = async () => {
+    const oracleLibraryFactory = await ethers.getContractFactory('OracleLibrary')
+    const oracleLibrary = await oracleLibraryFactory.deploy()
+    const oracleTestFactory = await ethers.getContractFactory('OracleTest', {
+      libraries: {
+        OracleLibrary: oracleLibrary.address,
+      },
+    })
+    const oracle = await oracleTestFactory.deploy()
+
+    await oracle.initialize({
+      time: 0,
+      tick: 0,
+      liquidity: 0,
+    })
+
+    await oracle.grow(4)
+    await oracle.update({ advanceTimeBy: 1, tick: 3, liquidity: 2 })
+    await oracle.update({ advanceTimeBy: 4, tick: -7, liquidity: 6 })
+    await oracle.update({ advanceTimeBy: 5, tick: -2, liquidity: 4 })
+
+    const tickMathTestFactory = await ethers.getContractFactory('TickMathTest')
+    const tickMath = await tickMathTestFactory.deploy()
+
+    return { oracle: oracle as OracleTest, tickMath: tickMath as TickMathTest }
+  }
+
+  describe('#consult', () => {
+    const sqrtRatioToRatioX128 = (x: BigNumber) => {
+      return x.pow(2).div(BigNumber.from(2).pow(64))
+    }
+
+    const getNearestTick = async (startTimestamp: BigNumber, endTimestamp: BigNumber) => {
+      const oracleTime = BigNumber.from(await oracle.time())
+      const tickCumulatives = (await oracle.observe([oracleTime.sub(startTimestamp), oracleTime.sub(endTimestamp)]))[
+        'tickCumulatives'
+      ]
+      const tick = tickCumulatives[1].sub(tickCumulatives[0]).div(endTimestamp.sub(startTimestamp))
+
+      return tick
+    }
+
+    it('returns correct price when tick cumulatives is increasing', async () => {
+      const startTimestamp = BigNumber.from(1)
+      const endTimestamp = BigNumber.from(6)
+
+      const tick = await getNearestTick(startTimestamp, endTimestamp)
+      const sqrtRatioX96 = await tickMath.getSqrtRatioAtTick(tick)
+
+      const calculatedPrice = await sqrtRatioToRatioX128(sqrtRatioX96).toString()
+      const oracleLibPrice = await oracle.consult(oracle.address, startTimestamp, endTimestamp)
+
+      expect(calculatedPrice).to.equal(oracleLibPrice)
+    })
+
+    it('returns correct price when tick cumulatives is decreasing', async () => {
+      const startTimestamp = BigNumber.from(3)
+      const endTimestamp = BigNumber.from(6)
+
+      const tick = await getNearestTick(startTimestamp, endTimestamp)
+      const sqrtRatioX96 = await tickMath.getSqrtRatioAtTick(tick)
+
+      const calculatedPrice = await sqrtRatioToRatioX128(sqrtRatioX96).toString()
+      const oracleLibPrice = await oracle.consult(oracle.address, startTimestamp, endTimestamp)
+
+      expect(calculatedPrice).to.equal(oracleLibPrice)
+    })
+
+    it('returns correct price when tick cumulatives is zeros', async () => {
+      const startTimestamp = BigNumber.from(0)
+      const endTimestamp = BigNumber.from(1)
+
+      const tick = await getNearestTick(startTimestamp, endTimestamp)
+      const sqrtRatioX96 = await tickMath.getSqrtRatioAtTick(tick)
+
+      const calculatedPrice = await sqrtRatioToRatioX128(sqrtRatioX96).toString()
+      const oracleLibPrice = await oracle.consult(oracle.address, startTimestamp, endTimestamp)
+
+      expect(calculatedPrice).to.equal(oracleLibPrice)
+    })
+
+    it('reverts when endTimestamp is not in observations', async () => {
+      const startTimestamp = BigNumber.from(6)
+      const endTimestamp = BigNumber.from(100000)
+
+      await expect(oracle.consult(oracle.address, startTimestamp, endTimestamp)).to.be.revertedWith('OLD')
+    })
+
+    it('reverts when endTimestamp is before startTimestamp', async () => {
+      const startTimestamp = BigNumber.from(6)
+      const endTimestamp = BigNumber.from(1)
+
+      await expect(oracle.consult(oracle.address, startTimestamp, endTimestamp)).to.be.revertedWith('Bad range')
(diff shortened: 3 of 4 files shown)


contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity =0.7.6;
+
+import '@uniswap/v3-core/contracts/libraries/FullMath.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@uniswap/v3-core/contracts/interfaces/pool/IUniswapV3PoolDerivedState.sol';
+import '@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol';
+import 'hardhat/console.sol';
```
moodysalem, line 8: leftover

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
...
+
+library OracleLibrary {
```
moodysalem, line 10: missing natspec https://docs.soliditylang.org/en/v0.7.6/natspec-format.html

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
...
+    function consult(
+        uint32 time,
```
moodysalem, line 12: we are pretty bad about this in core, but this should probably be `blockTimestampCurrent` instead of just `time`

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
...
+    ) public view returns (uint256 ratioX128) {
+        require(startBlockTimestamp < endBlockTimestamp, 'Bad range');
+        IUniswapV3PoolDerivedState oracle = IUniswapV3PoolDerivedState(poolAddress);
+
+        uint32[] memory secondAgos = new uint32[](2);
+        secondAgos[0] = time - startBlockTimestamp;
+        secondAgos[1] = time - endBlockTimestamp;
+
+        (int56[] memory tickCumulatives, ) = oracle.observe(secondAgos);
+        int24 tick = int24((tickCumulatives[1] - tickCumulatives[0]) / int56(endBlockTimestamp - startBlockTimestamp));
```
moodysalem, line 25: this division does not round consistently if tickCumulatives diff is negative or positive

recommend always rounding towards negative infinity

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
...
+import 'hardhat/console.sol';
+
+library OracleLibrary {
+    function consult(
+        uint32 time,
+        address poolAddress,
+        uint32 startBlockTimestamp,
+        uint32 endBlockTimestamp
+    ) public view returns (uint256 ratioX128) {
+        require(startBlockTimestamp < endBlockTimestamp, 'Bad range');
```
moodysalem, line 17: since start and end block timestamps are uint32s, this can be false in cases where you want the cumulative tick across an overflow boundary

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,30 @@
...
+import '@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol';
+import 'hardhat/console.sol';
+
+library OracleLibrary {
+    function consult(
+        uint32 time,
+        address poolAddress,
+        uint32 startBlockTimestamp,
+        uint32 endBlockTimestamp
+    ) public view returns (uint256 ratioX128) {
```
moodysalem, line 16: the ratio is a bit hard to work with, in previous examples we had the consult method take an amount of one token and return the amount of the other token, which is great because it lets the caller choose their denominator (instead of using `2**128`)

contracts/test/OracleTest.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.7.6;
+pragma abicoder v2;
+
+import '@uniswap/v3-core/contracts/libraries/Oracle.sol';
+import '../libraries/OracleLibrary.sol';
+
+contract OracleTest {
```
moodysalem, line 8: i'd rather not copy this in if we can avoid it by just testing directly against a pool with some helper functions for manipulating the price--one reason to do this is it would be more indicative of correctness of the oracle library

contracts/libraries/OracleLibrary.sol
moodysalem, line 16: just make this method internal and you don't have to deploy the library separately because the bytecode will get inlined in the caller contract

moodysalem (commented):
some initial thoughts

contracts/test/OracleTest.sol
smsunarto: Would using https://github.com/Uniswap/uniswap-v3-core/blob/main/contracts/test/MockTimeUniswapV3Pool.sol work? Unfortunately, I don't think it's exposed via NPM, and would still require copy and pasting the contract.

moodysalem: oh, yeah i see, it depends on mock time. let me think about it some more, this is an ok approach for now.

moodysalem:
also might consider doing some kind of rudimentary liquidity weighted version across pools for a pair

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+
+import '@uniswap/v3-core/contracts/libraries/FullMath.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol';
+
+/// @title Oracle library
+/// @notice Provides functions to compute time weighted token conversion rate using the oracle
+library OracleLibrary {
+    uint32 constant UINT32_MAX = uint32(-1);
```
moodysalem, line 12: you don't need a constant for this, just use `type(uint32).max`

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+    /// @param poolAddress The address of the pool to observe
+    /// @param tokenIn The address of the ERC20 token contract of amountIn
+    /// @param amountIn The token input amount before conversion
+    /// @param startBlockTimestamp The start of the time range to be observed
+    /// @param endBlockTimestamp The end of the time range to be observed
+    /// @return amountOut The token output amount after conversion
+    function consult(
+        uint32 currentBlockTimestamp,
+        address poolAddress,
+        address tokenIn,
```
moodysalem, line 26: there's nothing going in, so i'd prefer using the names `base` and `quote` instead of `in` and `out`

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+        uint256 amountIn,
+        uint32 startBlockTimestamp,
+        uint32 endBlockTimestamp
+    ) internal view returns (uint256 amountOut) {
+        require(startBlockTimestamp != endBlockTimestamp, 'Bad range');
+        IUniswapV3Pool oracle = IUniswapV3Pool(poolAddress);
+        require(tokenIn == oracle.token0() || tokenIn == oracle.token1(), 'Invalid token');
```
moodysalem, line 33: these two calls are going to be expensive. let's add gas tests. and maybe consider just taking token0, token1, fee as arguments and computing the pool address?

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+    /// @param amountIn The token input amount before conversion
+    /// @param startBlockTimestamp The start of the time range to be observed
+    /// @param endBlockTimestamp The end of the time range to be observed
+    /// @return amountOut The token output amount after conversion
+    function consult(
+        uint32 currentBlockTimestamp,
+        address poolAddress,
+        address tokenIn,
+        uint256 amountIn,
+        uint32 startBlockTimestamp,
```
moodysalem, line 28: i would consider taking a single `uint32 period` instead of start/end since that is the common case, and avoids all the edges with it

otherwise, if you want to take start/end, i would just take uint256 timestamps so you can just check that they are ordered, less than current timestamp, and the start timestamp is less than `2**32` seconds ago

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+        require(tokenIn == oracle.token0() || tokenIn == oracle.token1(), 'Invalid token');
+
+        uint32[] memory secondAgos = new uint32[](2);
+        secondAgos[0] = currentBlockTimestamp >= startBlockTimestamp
+            ? currentBlockTimestamp - startBlockTimestamp
+            : currentBlockTimestamp + (UINT32_MAX - startBlockTimestamp);
+
+        secondAgos[1] = currentBlockTimestamp >= endBlockTimestamp
+            ? currentBlockTimestamp - endBlockTimestamp
+            : currentBlockTimestamp + (UINT32_MAX - endBlockTimestamp);
```
moodysalem, line 42: this logic is confusing. see above comment about taking either `period` and going from `0 seconds ago to period seconds ago`, or taking larger types for start/end timestamp

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+        int24 tick = int24(tickCumulativesDelta / observationTimeDelta);
+
+        // Always round to negative infinity
+        tick = (tick < 0 && (tickCumulativesDelta % observationTimeDelta != 0)) ? tick - 1 : tick;
+
+        uint160 sqrtRatioX96 = TickMath.getSqrtRatioAtTick(tick);
+        uint256 ratioX128 = FullMath.mulDiv(sqrtRatioX96, sqrtRatioX96, 1 << 64);
+
+        if (tokenIn == oracle.token0()) {
```
moodysalem, line 59: another of the calls we want to avoid

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,65 @@
...
+                ? int56(endBlockTimestamp - startBlockTimestamp)
+                : int56(endBlockTimestamp + (UINT32_MAX - startBlockTimestamp));
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+        int24 tick = int24(tickCumulativesDelta / observationTimeDelta);
+
+        // Always round to negative infinity
+        tick = (tick < 0 && (tickCumulativesDelta % observationTimeDelta != 0)) ? tick - 1 : tick;
+
+        uint160 sqrtRatioX96 = TickMath.getSqrtRatioAtTick(tick);
+        uint256 ratioX128 = FullMath.mulDiv(sqrtRatioX96, sqrtRatioX96, 1 << 64);
```
moodysalem, line 57: what if sqrtRatioX96 is less than 128 bits (i.e. the sqrt price is less than 2^32, which is most the time)? can we be more precise by using a X192?

contracts/libraries/OracleLibrary.sol
smsunarto: hm, wouldn't that mean that mean we can't use the mock time Oracle anymore?

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,52 @@
...
+        (int56[] memory tickCumulatives, ) = oracle.observe(secondAgos);
+
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+        int24 tick = int24(tickCumulativesDelta / period);
+
+        // Always round to negative infinity
+        tick = (tick < 0 && (tickCumulativesDelta % period != 0)) ? tick - 1 : tick;
+
+        uint160 sqrtRatioX96 = TickMath.getSqrtRatioAtTick(tick);
+        uint256 ratioX192 = FullMath.mulDiv(sqrtRatioX96, sqrtRatioX96, 1);
```
moodysalem, line 46: this is equivalent to just `sqrtRatioX192*sqrtRatioX192`

it can overflow, for example, if the time weighted average tick is `TickMath.MAX_TICK`

we need to do something differently if it's less than or equal to `type(uint128).max` vs. greater than that number

contracts/test/OracleTest.sol
```diff
@@ -0,0 +1,31 @@
...
+        address baseToken,
+        address quoteToken,
+        uint24 fee,
+        uint256 baseAmount,
+        uint32 period
+    ) public view returns (uint256 quoteAmount) {
+        quoteAmount = OracleLibrary.consult(factory, baseToken, quoteToken, fee, baseAmount, period);
+    }
+
+    // For gas snapshot test
```
moodysalem, line 20: use .estimateGas or write a getGasCostOf method

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
```
moodysalem, line 2: my bad we should make this <0.8.0 as well because in 0.8.x math is safe by default

test/__snapshots__/OracleLibrary.gas.spec.ts.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`OracleLibrary gas tests #consult base token: token 0 1`] = `74764`;
```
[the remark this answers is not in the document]
smsunarto: Yeap, it's for the gas snapshot. Unless we don't want them anymore?

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,71 @@
...
+        secondAgos[0] = period;
+        secondAgos[1] = 0;
+
+        (int56[] memory tickCumulatives, ) = oracle.observe(secondAgos);
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+
+        tick = int24(tickCumulativesDelta / period);
+
+        // Always round to negative infinity
+        tick = (tick < 0 && (tickCumulativesDelta % period != 0)) ? tick - 1 : tick;
```
moodysalem, line 41: could you maybe avoid an op by doing `if (tick < 0 && (tickCumulativesDelta % period != 0)) tick--;`

moodysalem: needs to check the numerator < 0, rather than the quotient

moodysalem (changes requested):

test/OracleLibrary.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      const quoteAmount = await oracle.getQuoteAtTick(
+        BigNumber.from(getMaxTick(TEST_POOL.feeAmount)),
+        expandTo18Decimals(1),
+        tokens[0].address,
+        tokens[1].address
+      )
+      expect(quoteAmount).to.equal(BigNumber.from('271106558174734753828546514948592044174000833692526031838'))
+    })
+
+    it('token1: returns correct value when sqrtRatioX96 > type(uint128).max', async () => {
```
moodysalem, line 266: what about flipping token0/token1

test/shared/poolUtilities.ts
```diff
@@ -0,0 +1,178 @@
...
+  positions: Position[]
+  swapTests: SwapTestCase[]
+}
+
+export type SwapTestCase = SwapExact0For1TestCase | SwapExact1For0TestCase
+
+export const MIN_SQRT_RATIO = BigNumber.from('4295128739')
+export const MAX_SQRT_RATIO = BigNumber.from('1461446703485210103287273052203988822378723970342')
+
+export function getCreate2Address(
```
moodysalem, line 43: already have this https://github.com/Uniswap/uniswap-v3-periphery/blob/main/test/shared/computePoolAddress.ts

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// @param fee Fee amount that correspond to the pool that we want to observe
+    /// @param period Number of seconds in the past to start calculating time-weighted average
+    /// @return tick The time-weighted average tick from (block.timestamp - period) to block.timestamp
+    function consult(
+        address factory,
+        address token0,
+        address token1,
+        uint24 fee,
+        uint32 period
+    ) internal view returns (int24 tick) {
```
moodysalem, line 26: tick could be better named, `timeWeightedAverageTick`

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,59 @@
...
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+        int24 tick = int24(tickCumulativesDelta / secondsAgo);
+
+        // Always round to negative infinity
+        tick = (tick < 0 && (tickCumulativesDelta % secondsAgo != 0)) ? tick - 1 : tick;
+
+        uint160 sqrtRatioX96 = TickMath.getSqrtRatioAtTick(tick);
+
+        if (sqrtRatioX96 <= type(uint128).max) {
+            uint256 ratioX192 = FullMath.mulDiv(sqrtRatioX96, sqrtRatioX96, 1);
```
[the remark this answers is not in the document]
NoahZinsmeister: oh i missed the condition nvm me

NoahZinsmeister (approved):
i'll let @moodysalem chime in here as well, as i know he's been tracking this a bit more closely recently, but wanted to give a thumbs up to the actual OracleLibrary itself. I think it's really elegant, and a great minimum viable contract :)

there are lots of extensions i can think of, like consulting over routes (instead of pools), aggregating over multiple fee tiers, aggregating over many historical periods, fetching data for slices of time in the past, etc. etc., but i think this is a strong base, we should get it merged!

test/OracleLibrary.spec.ts
```diff
@@ -233,44 +233,48 @@ describe('OracleLibrary', () => {
...
+        BigNumber.from('92252781303487840575489571037701809782439021325602413126085498571520800050278')
+      )
     })
 
-    it('token0: returns correct value when sqrtRatioX96 > type(uint128).max', async () => {
+    it('token1: returns correct value when at max tick | sqrtRatioX96 > type(uint128).max', async () => {
       const quoteAmount = await oracle.getQuoteAtTick(
-        BigNumber.from(443637),
-        expandTo18Decimals(10000000),
+        BigNumber.from(getMaxTick(TEST_POOL.feeAmount)),
```
moodysalem, line 272: what about absolute max tick

moodysalem (approved):

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,71 @@
...
+library OracleLibrary {
+    /// @notice Fetches time-weighted average tick using Uniswap V3 oracle
+    /// @param factory Address of Uniswap V3 factory contract
+    /// @param token0 Address of an ERC20 token contract
+    /// @param token1 Address of an ERC20 token contract
+    /// @param fee Fee amount that correspond to the pool that we want to observe
+    /// @param period Number of seconds in the past to start calculating time-weighted average
+    /// @return timeWeightedAverageTick The time-weighted average tick from (block.timestamp - period) to block.timestamp
+    function consult(
+        address factory,
```
moodysalem, line 21: wait.. the reason we changed this was because we had to query token0, token1, but now it just runs the twap tick, so you can just take the pool

moodysalem: it can also greatly simplify testing if you just write a mock pool that only implements `#observe`

contracts/test/OracleTest.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.7.6;
+
+import '../libraries/OracleLibrary.sol';
+
+contract OracleTest {
+    uint32[2] private POSITIVE_TICK_CASE = [uint32(11), uint32(0)];
```
moodysalem, line 7: weird to hard code these test cases. ideally you write a separate contract that can be constructed to return values for any test case. then you dont touch the smart contracts while you work on unit tests

test/__snapshots__/OracleLibrary.spec.ts.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`OracleLibrary #consult gas test 1`] = `6129`;
```
moodysalem, line 3: that's ok this isnt realistic. we know what observe costs from core tests so this tells us the overhead of the code in the library

contracts/libraries/OracleLibrary.sol
```diff
@@ -0,0 +1,61 @@
...
+/// @title Oracle library
+/// @notice Provides functions to integrate with V3 pool oracle
+library OracleLibrary {
+    /// @notice Fetches time-weighted average tick using Uniswap V3 oracle
+    /// @param pool Address of Uniswap V3 pool that we want to observe
+    /// @param period Number of seconds in the past to start calculating time-weighted average
+    /// @return timeWeightedAverageTick The time-weighted average tick from (block.timestamp - period) to block.timestamp
+    function consult(address pool, uint32 period) internal view returns (int24 timeWeightedAverageTick) {
+        require(period != 0, 'BP');
+        IUniswapV3Pool oracle = IUniswapV3Pool(pool);
```
moodysalem, line 19: dont mstore this, just do IUniswapV3Pool(pool).observe

moodysalem (changes requested):
final 2 changes

test/OracleLibrary.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    const tokenFactory = await ethers.getContractFactory('TestERC20')
+    const tokens = (await Promise.all([
+      tokenFactory.deploy(constants.MaxUint256.div(2)), // do not use maxu256 to avoid overflowing
+      tokenFactory.deploy(constants.MaxUint256.div(2)),
+    ])) as [TestERC20, TestERC20]
+
+    tokens.sort((a, b) => (a.address.toLowerCase() < b.address.toLowerCase() ? -1 : 1))
+
+    const oracleFactory = await ethers.getContractFactory('OracleTest')
+    const oracles = (await Promise.all([
```
moodysalem, line 29: per the above, you don't need to construct all your mocks in the fixture

moodysalem (approved):

moodysalem (approved):
at this point, suggestions optional. lgtm.

test/OracleLibrary.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      const oracleTick = await oracle.consult(mockObservable.address, period)
+
+      // Always round to negative infinity
+      // In this case, we need to subtract one because integer division rounds to 0
+      expect(oracleTick).to.equal(BigNumber.from(-1))
+    })
+
+    it('correct output for tick cumulatives across overflow boundaries', async () => {
+      const period = 4
+      const tickCumulatives = [BigNumber.from(-100), BigNumber.from('36028797018963967')]
```
moodysalem, line 98: how does this come out to 24?

test/OracleLibrary.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      const oracleTick = await oracle.consult(mockObservable.address, period)
+
+      // Always round to negative infinity
+      // In this case, we don't have do anything
+      expect(oracleTick).to.equal(BigNumber.from(24))
+    })
+
+    it('correct output for tick cumulatives across underflow boundaries', async () => {
+      const period = 4
+      const tickCumulatives = [BigNumber.from(100), BigNumber.from('-36028797018963967')]
```
moodysalem, line 109: same how is this -25

moodysalem (approved):

PR #104, opened 2021-04-21
NFTDescriptor: escape quotes

ewilz:
closes https://github.com/Uniswap/uniswap-v3-periphery/issues/56

### DIFF
--- a/contracts/libraries/NFTDescriptor.sol
+++ b/contracts/libraries/NFTDescriptor.sol
@@ -113,11 +113,12 @@ library NFTDescriptor {
     }
 
     function escapeQuotes(string memory symbol) internal pure returns (string memory) {
-      bytes memory symbolBytes = bytes(symbol);
-      uint8 quotesCount = 0;
-      for(uint8 i = 0; i < symbolBytes.length; i++) {
-        if (symbolBytes[i] == '"') {
-          quotesCount++;
+        bytes memory symbolBytes = bytes(symbol);
+        uint8 quotesCount = 0;
+        for (uint8 i = 0; i < symbolBytes.length; i++) {
+            if (symbolBytes[i] == '"') {
+                quotesCount++;
+            }
         }
         if (quotesCount > 0) {
             bytes memory formattedBytes = new bytes(symbolBytes.length + (quotesCount));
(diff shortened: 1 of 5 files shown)


contracts/libraries/NFTDescriptor.sol
```diff
@@ -110,22 +112,36 @@ library NFTDescriptor {
...
+                }
+                formattedBytes[index++] = symbolBytes[i];
+            }
+            return string(formattedBytes);
+        }
+        return symbol;
+    }
+
     struct DecimalStringParams {
-        // significant figures of decimal
```
ewilz, line 114: prettier's gone rogue...

contracts/libraries/NFTDescriptor.sol
```diff
@@ -113,43 +113,35 @@ library NFTDescriptor {
     }
 
     function formatTokenSymbol(string memory symbol) internal pure returns (string memory) {
```
moodysalem, line 115: seems this method should be more like `escapeQuotes(string) returns (string)` and could use some fuzz testing (it may not work right multibyte characters, not sure)

contracts/libraries/NFTDescriptor.sol
fvictorio: Ugh, I'm so sorry about this. I just opened a PR that fixes this: https://github.com/prettier-solidity/prettier-plugin-solidity/pull/480 I'll try to have it merged and released today.

ewilz: Wow, Franco, that was so quick!! thank you! cc @fvictorio

fvictorio: It should be fixed in `prettier-plugin-solidity@1.0.0-beta.10`. Let us know if it's not!

ewilz: yess, that did the trick 💯 happy to be able to throw you weird edgecases as you guys get closer to a stable release 😂

PaulRBerg:
Interesting edge case!

I'm curious: how many ERC-20 tokens out there have quotes in their symbols (is there any particular example you could share)? Shouldn't this be very rare?

PR #109, opened 2021-04-27
Optimize base64 a bit

ZumZoom:


moodysalem:
Thanks @ZumZoom! I just synced to Brecht's latest commit and it looks like it's a bit cheaper than this (and includes most the optimizations)

PR #112, opened 2021-04-27
Include Shippable NFT Image + Base64 All metadata

ewilz:


test/NFTDescriptor.spec.ts
```diff
@@ -609,56 +647,202 @@ describe('NFTDescriptor', () => {
...
+    const x1 = scale(tokenToNum(quoteTokenAddress, 36), 0, 255, 16, 274)
+    const y1 = scale(tokenToNum(baseTokenAddress, 36), 0, 255, 100, 484)
+    const x2 = scale(tokenToNum(quoteTokenAddress, 32), 0, 255, 16, 274)
+    const y2 = scale(tokenToNum(baseTokenAddress, 32), 0, 255, 100, 484)
+    const x3 = scale(tokenToNum(quoteTokenAddress, 28), 0, 255, 16, 274)
+    const y3 = scale(tokenToNum(baseTokenAddress, 28), 0, 255, 100, 484)
+    const str1length = 7 * (tokenId.toString().length + 4 + 4)
+    const str2length = 7 * (tickLower.toString().length + 5 + 4)
+    const str3length = 7 * (tickUpper.toString().length + 5 + 4)
+    return `<svg width="290" height="500" viewBox="0 0 290 500" xmlns="http://www.w3.org/2000/svg" xmlns:xlink='http://www.w3.org/1999/xlink'><style>@import \
```
ewilz, line 802: soo....I did this. Was it the best way to do this?? I have no idea.

contracts/libraries/NFTDescriptor.sol
```diff
@@ -34,80 +36,57 @@ library NFTDescriptor {
...
         return
             string(
                 abi.encodePacked(
-                    'data:application/json,{"name":"',
+                    'data:application/json,',
+                    '{"name":"',
                     name,
                     '", "description":"',
-                    description,
+                    descriptionPartOne,
```
ewilz, line 85: THIS description in two parts was my Stack Too Deep error...of all things

moodysalem (changes requested):

test/NFTDescriptor.spec.ts
ewilz: yes you're right. this is actually a shit show 😝. I do need to test some of the text but I can parse the little snippets of the JSON that i Need to test.

moodysalem: if you want to validate the svg, i would parse the svg and check its elements

PR #117, opened 2021-04-29
SVG touchups

ewilz:
* Center Top Curve Point
* Get rid of illuminated Border Corners
* Clarify Max _Ticks_: Min _Ticks_: in svg
* Switch direction of fade (was fading opposite of what it should be)
* Curve Depth considers tickSpacing and has tighter thresholds
* Generalized test cleanup


contracts/libraries/NFTSVG.sol
```diff
@@ -238,7 +238,7 @@ library NFTSVG {
             curve = curve4;
         } else if (tickRange <= 100) {
             curve = curve5;
-        } else if (tickRange <= 10_000) {
+        } else if (tickRange <= 1_000) {
```
moodysalem, line 241: i just realized, we should look at the `(tickUpper - tickLower) / tickSpacing`, rather than the absolute difference, because it depends on the tick spacing. 5 ticks difference is not possible in a 200 tick spacing pool

ewilz: Does tick spacing effect the price at each tick (in absolute terms?)

moodysalem: also, we may want to use powers of 2 instead of powers of 10, 8 curves with powers of 2 is 256, and 256 tick spacings is == 167x price diff with high fee and 1.29 with low fee

moodysalem (approved):

PR #122, opened 2021-05-03
PermitRegistry.sol

albertocuestacanada:
A Lens contract that allows contracts to be registered with the off-chain permit they use, and with the parameters to build their domain separator.


moodysalem:
Will finish this up on the local branch

PR #124, opened 2021-05-03
✨ super rare special sparkle ✨

ewilz:
Best PR name evr  --^


contracts/libraries/NFTSVG.sol
```diff
@@ -369,4 +374,62 @@ library NFTSVG {
...
+                    '23L10.2181 13.7819L1.6077 18L9.56587 12.6522L0 12L9.56587 11.3478L1.6077 6L10.2181 10.2181L6 1.6077L11.3478 9.56587L12 0Z" fill="white"/></g>'
+                )
+            );
+        } else {
+            svg = '';
+        }
+    }
+
+    function isRare(uint256 tokenId) internal pure returns (bool) {
+        return hashAndCondense(tokenId) < 0x80000000 / logTwo(tokenId);
```
moodysalem, line 394: why not just use type(uint256).max / msb(tokenId) rather than "condense" it?

contracts/libraries/NFTSVG.sol
```diff
@@ -54,7 +54,12 @@ library NFTSVG {
                     ),
                     generateSVGCardMantle(params.quoteTokenSymbol, params.baseTokenSymbol, params.feeTier),
                     generageSvgCurve(params.tickLower, params.tickUpper, params.tickSpacing, params.overRange),
-                    generateSVGPositionDataAndLocationCurve(params.tokenId, params.tickLower, params.tickUpper),
+                    generateSVGPositionDataAndLocationCurve(
+                        params.tokenId.toString(),
```
moodysalem, line 58: probably should do this toString in the method

contracts/libraries/NFTSVG.sol
```diff
@@ -369,4 +374,62 @@ library NFTSVG {
...
+    function isRare(uint256 tokenId) internal pure returns (bool) {
+        return hashAndCondense(tokenId) < 0x80000000 / logTwo(tokenId);
+    }
+
+    function hashAndCondense(uint256 tokenId) internal pure returns (uint256) {
+        bytes32 h = sha256(abi.encodePacked(tokenId)) & bytes32(uint256(0xffffffff));
+        return uint256(h);
+    }
+
+    function logTwo(uint256 tokenId) internal pure returns (uint256) {
```
moodysalem, line 402: delete and use BitMath

moodysalem (changes requested):

contracts/libraries/NFTSVG.sol
ewilz: I use it twice in the function, so it's nice to just have it as a string right in the calldata?

contracts/libraries/NFTSVG.sol
```diff
@@ -369,4 +376,28 @@ library NFTSVG {
...
+            svg = '';
+        }
+    }
+
+    function isRare(uint256 tokenId, address poolAddress) internal pure returns (bool) {
+        return getHash(tokenId, poolAddress) < type(uint256).max / (1 + BitMath.mostSignificantBit(tokenId) * 2);
+    }
+
+    function getHash(uint256 tokenId, address poolAddress) internal pure returns (uint256) {
+        bytes32 h = sha256(abi.encodePacked(tokenId, poolAddress));
```
moodysalem, line 400: is this different from keccak256? any reason to use one or the other?

ewilz: nope, updated though. saved gas

moodysalem: this func is tiny and used in one place, does it make sense to inline? also don't see the hash function update is pushed

contracts/libraries/NFTSVG.sol
moodysalem: nbd, but if we were going to call this onchain it would be cheaper to pass a uint and then store the converted string in a local variable

moodysalem (approved):

contracts/libraries/NFTSVG.sol
ewilz: it brought gas up by a whopping 13 on the view.

contracts/libraries/NFTSVG.sol
ewilz: you must be looking at a commit?  (these comments are listed as outdated) Will do, though

PR #146, opened 2021-06-14
feat: weighted oracle library

### LINKED ISSUE #140: Weighted Oracle library
Create an opinionated library (a la https://github.com/Uniswap/uniswap-v3-periphery/issues/22) for getting an time weighted average tick with some kind of weighting across fee tiers (e.g. liquidity)

nchamo:
Hey Uniswap team!

This PR is an attempt at https://github.com/Uniswap/uniswap-v3-periphery/issues/140. Each tier's weight is based on its liquidity, but I'm not using exactly the time-weighted liquidity. If you think there is a better way to calculate the weight, please let me know and I'll do it.

I wasn't sure if I should add this functionality to the previous oracle library (https://github.com/Uniswap/uniswap-v3-periphery/pull/96), or if the idea of the issue was to create a new library by itself. I went with the second option, but I can move it if necessary.

Also, I still have to test `consultPool`, but I wanted your feedback on the comment above, because the tick tests would be exactly the same as the other library. 

Let me know what you think! 😄 

closes #140


nchamo:
Sorry for the tag @moodysalem . This isn't urgent or anything, but I just wanted to make sure the PR isn't overlooked because I can't add reviewers.

nchamo:
Hey @moodysalem , @NoahZinsmeister !

Any chance you could take a look at this?

Validating the weighting approach would be more than enough to unblock me if you are too busy to do a full review 😄

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,53 @@
...
+
+        uint32[] memory secondsAgos = new uint32[](2);
+        secondsAgos[0] = period;
+        secondsAgos[1] = 0;
+        uint160 periodX128 = uint160(period) << 128;
+
+        uint136 liquidityWeightAccumulator;
+        uint152 liquidityWeightedTickAccumulator;
+
+        for (uint8 i = 0; i < pools.length; i++) {
```
moodysalem, line 25: make sure `pools.length <= type(uint8).max`

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,53 @@
...
+    function consultPool(address pool, uint32 period, uint32[] memory secondsAgos, uint160 periodX128) internal view returns (int24 timeWeightedAverageTick, uint128 liquidityWeight) {
+        (int56[] memory tickCumulatives, uint160[] memory secondsPerLiquidityCumulativeX128s) = IUniswapV3Pool(pool).observe(secondsAgos);
+        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+        uint160 secondsPerLiquidityCumulativesDelta = secondsPerLiquidityCumulativeX128s[1] - secondsPerLiquidityCumulativeX128s[0];
+
+        timeWeightedAverageTick = int24(tickCumulativesDelta / period);
+        if (tickCumulativesDelta < 0 && (tickCumulativesDelta % period != 0)) timeWeightedAverageTick--;
+
+        // This value isn't the time-weighted liquidity, but it can still helps us weight the tick based on tier liquidity
+        liquidityWeight = uint128(periodX128 / secondsPerLiquidityCumulativesDelta);
```
moodysalem, line 51: here's an example of computing time weighted harmonic mean liquidity for the period
https://github.com/Uniswap/uniswap-v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/contracts/test/OracleEchidnaTest.sol#L65-L68

once you have that, you can simply do an arithmetic mean weighted tick average, or pick the highest, or use the median

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,53 @@
...
+    /// @param pools Addresses of different tiers of the same Uniswap V3 pair that we want to observe
+    /// @param period Number of seconds in the past to start calculating liquidity-time-weighted average
+    /// @return liquidityTimeWeightedAverageTick The liquidity-time-weighted average tick from (block.timestamp - period) to block.timestamp
+    function consult(address[] calldata pools, uint32 period) internal view returns (int24 liquidityTimeWeightedAverageTick) {
+        require(period != 0, 'BP');
+
+        uint32[] memory secondsAgos = new uint32[](2);
+        secondsAgos[0] = period;
+        secondsAgos[1] = 0;
+        uint160 periodX128 = uint160(period) << 128;
```
moodysalem, line 20: computing periodX128 and then passing into inner function is weird when you already pass period and this is a simple shift

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,53 @@
...
+
+        uint136 liquidityWeightAccumulator;
+        uint152 liquidityWeightedTickAccumulator;
```
moodysalem, line 23: 128 + 8 + 24 = 160

could use a comment how you came up with these widths

contracts/libraries/WeightedOracleLibrary.sol
moodysalem: further how you do this maybe should be up to the caller? an enum? or have separate methods?

moodysalem:
btw thank you for looking into this! was on vacation last week, sorry for the delayed review

contracts/libraries/WeightedOracleLibrary.sol
nchamo: I really like that idea!

I split the methods and added an arithmetic mean weighted tick average like you suggested, so there is at least one default implementation

nchamo:
Thanks for the review @moodysalem !

Just implemented your suggestions 😄

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+    struct TimeWeightedObservation {
+        int24 timeWeightedAverageTick;
+        uint256 timeWeightedHarmonicMeanLiquidity;
+    }
+
+    /// @notice Fetches time-weighted observations across different tiers of the same Uniswap V3 pair
+    /// @param pools Addresses of different tiers of the same Uniswap V3 pair that we want to observe
+    /// @param period Number of seconds in the past to start calculating the time-weighted observations
+    /// @return observations An array of obervations that have been time-weighted from (block.timestamp - period) to block.timestamp
+    function consult(address[] calldata pools, uint32 period) internal view returns (TimeWeightedObservation[] memory observations) {
```
moodysalem, line 19: there is no guarantee in the interface these pools are for the same token pair

you may want to take `address token0, address token1, uint24[] feeAmounts` and compute pool addresses to enforce it, but that unfortunately also requires you to pass in factory address

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+
+/// @title Weighted Oracle library
+/// @notice Provides functions to integrate with different tier oracles of the same V3 pair
+library WeightedOracleLibrary {
+
+    struct TimeWeightedObservation {
+        int24 timeWeightedAverageTick;
+        uint256 timeWeightedHarmonicMeanLiquidity;
```
moodysalem, line 12: using "mean" here and "average" for tick, maybe consistently use one or other. average is more human readable

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+
+    /// @notice Fetches time-weighted observations across different tiers of the same Uniswap V3 pair
+    /// @param pools Addresses of different tiers of the same Uniswap V3 pair that we want to observe
+    /// @param period Number of seconds in the past to start calculating the time-weighted observations
+    /// @return observations An array of obervations that have been time-weighted from (block.timestamp - period) to block.timestamp
+    function consult(address[] calldata pools, uint32 period) internal view returns (TimeWeightedObservation[] memory observations) {
+        require(period != 0, 'BP');
+
+        observations = new TimeWeightedObservation[](pools.length);
+        uint256 periodX160 = uint256(period) * type(uint160).max;
```
moodysalem, line 23: you can just shift 160 since it's a uint256

contracts/libraries/WeightedOracleLibrary.sol
moodysalem: also i think this should be a uint128

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+
+        for (uint256 i; i < pools.length; i++) {
+          (int56[] memory tickCumulatives, uint160[] memory secondsPerLiquidityCumulativeX128s) = IUniswapV3Pool(pools[i]).observe(secondsAgos);
+          int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
+          uint160 secondsPerLiquidityCumulativesDelta = secondsPerLiquidityCumulativeX128s[1] - secondsPerLiquidityCumulativeX128s[0];
+
+          observations[i].timeWeightedAverageTick = int24(tickCumulativesDelta / period);
+          if (tickCumulativesDelta < 0 && (tickCumulativesDelta % period != 0)) observations[i].timeWeightedAverageTick--;
+
+          observations[i].timeWeightedHarmonicMeanLiquidity = periodX160 / (uint256(secondsPerLiquidityCumulativesDelta) << 32);
```
moodysalem, line 37: why are we shifting the denominator left 32? can't we just use `periodX128 / secondsPerLiquidityCumulativesDelta` and get the exact same precision?

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    /// @notice Given some time-weighted observations, calculates the arithmetic mean weighted tick average
+    /// @param observations A list of time-weighted observations
+    /// @return arithmeticMeanWeightedTick The arithmetic mean average tick, weighted by the observations' time-weighted harmonic mean liquidity
+    function getArithmeticMeanWeightedTick(TimeWeightedObservation[] memory observations) internal pure returns (int24 arithmeticMeanWeightedTick) {
```
moodysalem, line 44: probably needs a better name to indicate it's not just the average tick, but the tick arithmetic-mean-weighted by liquidity

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,56 @@
...
+
+        int256 weightedTicksAccumulator;
+        uint256 liquidityWeightAccumulator;
```
moodysalem, line 47: can we add comments describing what they accumulate and why they will never overflow?

contracts/libraries/WeightedOracleLibrary.sol
moodysalem: i know this makes testing harder, but the natspec indicates that it returns data `across different tiers of the same Uniswap V3 pair`. you can also clarify the natspec

contracts/libraries/WeightedOracleLibrary.sol
nchamo: That was my initial approach, but I changed it based on the link you provided
https://github.com/Uniswap/uniswap-v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/contracts/test/OracleEchidnaTest.sol#L65-L68

But I'll change it back, no problem.

moodysalem: oops i don't remember why i did that, never read the code i linked 😃 
there might be a reason, let me circle back

contracts/libraries/WeightedOracleLibrary.sol
nchamo: You are right. Now that this function only returns observations, there is no need to have the pools be tiers of the same pair. I'll update the natspec

contracts/libraries/WeightedOracleLibrary.sol
nchamo: I don't fully understand why using periodX128 without the `secondsPerLiquidityCumulativesDelta` shifting could lead to the number exceeding `type(uint128).max`. I agree that I don't want it to exceed it, so I'll go with whatever approach makes it happen. I just thought that this would also achieve it:
```solidity
observations[i].timeWeightedHarmonicAverageLiquidity = periodX128 / secondsPerLiquidityCumulativesDelta;
```

moodysalem: if period is 1 second, periodX128 is `1<<128`, and if liquidity is `type(uint128).max`, `secondsPerLiquidityCumulativesDelta` increases by `1<<128 / type(uint128).max == 1`, so you end up with `1<<128 / 1` which is g.t. `type(uint128).max`

nchamo: Yeah, just run into that same problem in one of my tests haha

Will keep the shifting approach then!

nchamo:
Sorry @moodysalem , forgot to let you know that I did the changes we discussed 😄

test/WeightedOracleLibrary.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+        secondsPerLiqCumulatives: secondsPerLiqCumulatives1
+      })
+
+      const mockObservable2 = await observableWith({
+        period,
+        tickCumulatives: [-7, -12],
+        secondsPerLiqCumulatives: secondsPerLiqCumulatives2
+      })
+
+      it('correct output for liquidity overflow', async () => {
```
NoahZinsmeister, line 101: is this `it` nested?

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+pragma solidity >=0.5.0 <0.8.0;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+
+/// @title Weighted Oracle library
+/// @notice Provides functions to integrate with different tier oracles of the same V3 pair
+library WeightedOracleLibrary {
+
+    struct TimeWeightedObservation {
+        int24 timeWeightedAverageTick;
```
NoahZinsmeister, line 11: i'd suggest `arithmeticMeanTick`, clearer than average, and imo `timeWeighted` is a confusing prefix, the name of the struct is good enough

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+        uint128 timeWeightedHarmonicAverageLiquidity;
```
NoahZinsmeister, line 12: i'd suggest `harmonicMeanLiquidity`

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// @notice Fetches time-weighted observations across different Uniswap V3 pools
```
NoahZinsmeister, line 15: i'd add a note indicating that pools should almost certainly be for the same underlying assets, and differ only in fee tier

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+    /// @param pools Addresses of different Uniswap V3 pools that we want to observe
+    /// @param period Number of seconds in the past to start calculating the time-weighted observations
+    /// @return observations An array of obervations that have been time-weighted from (block.timestamp - period) to block.timestamp
+    function consult(address[] calldata pools, uint32 period) internal view returns (TimeWeightedObservation[] memory observations) {
+        require(period != 0, 'BP');
+
+        observations = new TimeWeightedObservation[](pools.length);
+        uint192 periodX160 = uint192(period) * type(uint160).max;
```
NoahZinsmeister, line 23:
```suggestion
        uint192 periodX160 = uint192(period) << 160;
```

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// @notice Given some time-weighted observations, calculates the arithmetic mean tick, weighted by liquidity
+    /// @param observations A list of time-weighted observations
+    /// @return arithmeticMeanWeightedTick The arithmetic mean tick, weighted by the observations' time-weighted harmonic average liquidity
+    function getArithmeticMeanTickWeightedByLiquidity(TimeWeightedObservation[] memory observations) internal pure returns (int24 arithmeticMeanWeightedTick) {
+
+        // Accumulates the sum of all observations' products between each their own average tick and harmonic average liquidity
+        // Each product can be stored in a int160, so it would take approximatelly 2**96 observations to overflow this accumulator
+        int256 weightedTicksAccumulator;
```
NoahZinsmeister, line 50: personally, i'd call this numerator

contracts/libraries/WeightedOracleLibrary.sol
NoahZinsmeister: ditto, denominator

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,63 @@
...
+        // Accumulates the sum of the harmonic average liquidities from the given observations
+        // Each average liquidity can be stored in a uint128, so it will take approximatelly 2**128 observations to overflow this accumulator
+        uint256 liquidityWeightAccumulator;
+
+        for (uint256 i; i < observations.length; i++) {
+            weightedTicksAccumulator += int256(observations[i].timeWeightedHarmonicAverageLiquidity) * observations[i].timeWeightedAverageTick;
+            liquidityWeightAccumulator += observations[i].timeWeightedHarmonicAverageLiquidity;
+        }
+
+        arithmeticMeanWeightedTick = int24(weightedTicksAccumulator / int256(liquidityWeightAccumulator));
```
NoahZinsmeister, line 61: might want to round down here as well?

test/WeightedOracleLibrary.spec.ts
nchamo: Good catch! My bad

contracts/libraries/WeightedOracleLibrary.sol
nchamo: Actually, we are using `type(uint160).max` on purpose, to make sure that the liquidity value doesn't overflow.

We discussed it with @moodysalem [here](https://github.com/Uniswap/uniswap-v3-periphery/pull/146#discussion_r656352449)

There is also a test that fails in that case 😄

contracts/libraries/WeightedOracleLibrary.sol
nchamo: Done!

Could you check if it is clear enough now?

contracts/libraries/WeightedOracleLibrary.sol
nchamo: Done!

nchamo:
Hey @moodysalem  and @NoahZinsmeister !

Are there any more changes you'd like me to make?

Also, I couldn't find any code/library that would determine how far back the `pool.observe` can be used without reverting. Is there any? If not, then I'll probably create another PR for that 😄

nchamo:
FYI @moodysalem , added `getOldestObservationTimestampForPool` to the library. It returns the oldest timestamp available on the pool. Did it because otherwise it would be really difficult to avoid reverts when calling `consult`

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,81 @@
...
+/// @title Weighted Oracle library
+/// @notice Provides functions to integrate with different tier oracles of the same V3 pair
+library WeightedOracleLibrary {
+
+    struct TimeWeightedObservation {
+        int24 arithmeticMeanTick;
+        uint128 harmonicMeanLiquidity;
+    }
+
+    /// @notice Fetches time-weighted observations across different Uniswap V3 pools. These pools should almost certainly be for the same underlying assets, and differ only in fee tier
```
moodysalem, line 15: there isn't much reason for this function to take an array of pools vs. one pool, i would probably change the signature actually and `TimeWeightedObservation` should be more like `PeriodObservation` (realize this is basically the same as OracleLibrary but returns harmonic mean liquidity as well)

might be good to add documentation to the struct as well

contracts/libraries/WeightedOracleLibrary.sol
```diff
@@ -0,0 +1,81 @@
...
+        arithmeticMeanWeightedTick = int24(numerator / int256(denominator));
+
+        // Always round to negative infinity
+        if (numerator < 0 && (numerator % int256(denominator) != 0)) arithmeticMeanWeightedTick--;
+    }
+
+    /// @notice Given a Uniswap V3 pool, returns the timestamp of the oldest observation available
+    /// @param pool Address of Uniswap V3 pool that we want to analyze
+    /// @return oldestBlockTimestamp The timestamp of the oldest observation available on the pool
+    function getOldestObservationTimestampForPool(address pool) internal view returns (uint32 oldestBlockTimestamp) {
```
moodysalem, line 70: it seems like you wouldn't ever want to sacrifice oracle security just because the data isn't available, but rather you'd just increase observation cardinality and then wait the required period for enough observations to be collected, so i would prefer not to expose this

nchamo: Ok, will revert then

nchamo:
@moodysalem , did the changes you requested 😄 

Also, the linter seems to be failing with the following error
```This action does not have permission to create annotations on forks.```

Is there anything I can do for it to run correctly?

**Comment by 404skillz** [comment]
Hey there guys, I'm looking into integrating this in a project to replace some oracles, any news ? @nchamo @moodysalem

**Comment by nchamo** [comment]
Hey @404skillz , sorry for the delay. I think there is nothing left on my side to do, I implemented all changes

**Review by NoahZinsmeister — APPROVED** [review]
LGTM, can you take another look @moodysalem, and then we can merge?

**Comment by NoahZinsmeister** [comment]
@nchamo sorry for the delay! was out last week, let's try to land this this week :)

**Comment by NoahZinsmeister** [comment]
hmm, actually i wish there were a better way to convey that `getArithmeticMeanTickWeightedByLiquidity` should almost always only be called with `observations` where `period` is the same and `pools` have identical underlying tokens

getting the average weighted price over different `period`s is obviously weird (though i guess conceivably reasonable in some setting?)

getting the average weighted price over `pool`s with non-identical underlying tokens is likely dangerous, because a) the prices are not in general going to be comparable, but b) even in the case where they could be (e.g. averaging over some combo of USDC<>ETH/DAI<>ETH pools to get the ETH price in USD), liquidity values are not guaranteed to be comparable unless the decimals of the USD token are also equal (there might be a scale transform we could apply in this case? not sure, would have to work out the math, cc @danrobinson if you're interested)

**Review by moodysalem — APPROVED** [review]

**Comment by nchamo** [comment]
> hmm, actually i wish there were a better way to convey that `getArithmeticMeanTickWeightedByLiquidity` should almost always only be called with `observations` where `period` is the same and `pools` have identical underlying tokens
> 
> getting the average weighted price over different `period`s is obviously weird (though i guess conceivably reasonable in some setting?)
> 
> getting the average weighted price over `pool`s with non-identical underlying tokens is likely dangerous, because a) the prices are not in general going to be comparable, but b) even in the case where they could be (e.g. averaging over some combo of USDC<>ETH/DAI<>ETH pools to get the ETH price in USD), liquidity values are not guaranteed to be comparable unless the decimals of the USD token are also equal (there might be a scale transform we could apply in this case? not sure, would have to work out the math, cc @danrobinson if you're interested)

@NoahZinsmeister I agree. Mixing periods or pools would almost surely return unwanted results.  I think there are a few possible approaches:
* Explain it better on the comments (the part about using the same pair is explained, but not the part about using the same period)
* Add an example to the docs, like https://github.com/Uniswap/uniswap-v3-periphery/pull/161
* Have `PeriodObservation` contain the period/tokens and make sure in `getArithmeticMeanTickWeightedByLiquidity` that they are all the same (the period part is easy, but getting the pool's tokens would require a few more calls to the pool)
* Add a new function to the library that takes a pair, some fee tiers and a period, and "does everything for you"

What do you think? Is that something that should be tackled here, on in another PR?

File: contracts/libraries/WeightedOracleLibrary.sol [hunk 20]
```diff
@@ -0,0 +1,64 @@
...
+        // Always round to negative infinity
+        if (tickCumulativesDelta < 0 && (tickCumulativesDelta % period != 0)) observation.arithmeticMeanTick--;
+
+        // We are shifting the liquidity delta to ensure that the result doesn't overflow uint128
+        observation.harmonicMeanLiquidity = uint128(periodX160 / (uint192(secondsPerLiquidityCumulativesDelta) << 32));
+    }
+
+    /// @notice Given some time-weighted observations, calculates the arithmetic mean tick, weighted by liquidity
+    /// @param observations A list of time-weighted observations
+    /// @return arithmeticMeanWeightedTick The arithmetic mean tick, weighted by the observations' time-weighted harmonic average liquidity
```
NoahZinsmeister, line 43:
```suggestion
    /// @return arithmeticMeanWeightedTick The arithmetic mean tick, weighted by the observations' time-weighted harmonic average liquidity
    /// @dev In most scenarios, each entry of `observations` should share the same `period` and underlying `pool` tokens.
    /// If `period` differs across observations, the result becomes difficult to interpret and is likely biased/manipulable.
    /// If the underlying `pool` tokens differ across observations, extreme care must be taken to ensure that both prices and liquidity values are comparable.
    /// Even if prices are commensurate (e.g. two different USD-stable assets against ETH), liquidity values may not be, as decimals can differ between tokens.
```

NoahZinsmeister:
@nchamo appreciate the quick response :)

i've taken the liberty of applying

> Explain it better on the comments

in https://github.com/Uniswap/uniswap-v3-periphery/pull/146/commits/695b3a92f343a084bfbccc68615977e5ead58151

i discussed offline with @moodysalem, and we felt that either verifying tokens or locking users into a single API were too restrictive

PR #147, opened 2021-06-16
feat: OVM

mds1:
## Summary

The OVM is a sandboxed environment for the EVM which allows execution of smart contract code on L2, with the ability to dispute it on L1 via a fraud proof. Solidity code can be compiled directly to OVM code instead of EVM code by using ["optimistic solc"](https://github.com/ethereum-optimism/solidity), a fork of the well-known solc. The OVM bytecode is ~30% larger than EVM bytecode, which may result in the bytecode being larger than the 24kb size limit for smart contracts.

This PR refactors the codebase to allow it to fit in the 24kb limit, and does a minimal diff to the test suite so that tests can pass when ran against an instance of [go-ethereum running the OVM](https://github.com/ethereum-optimism/optimism/tree/develop/l2geth). We suspect the test suite changes around waiting for txs to be confirmed would've been required anyway if we wanted to test against geth or any full node impl.

## Explanation of code changes

- Main contract changes:
    - EOA `permit` support was removed since all OVM wallets are EIP-1271 compatible contract wallets
    - `getPopulatedTicksInWord` uses too much gas so `getPopulatedTicksInPartialWord` was added
    - Always use 1 run of the optimizer to minimize contract size`
    - Changes to reduce size of `NonfungiblePositionManager` below 24 kb include:
        - Extract a library of functions in `NonfungiblePositionLibrary` and out of `NonfungiblePositionManager`.
        - WETH address is now a constant in `NonfungiblePositionLibrary` but still parameterized elsewhere
        - Remove the explicit `NonfungiblePositionManager.positions()` function and use the getter generated on a public `positions` mapping
        - Merge `NonfungibleTokenPositionDescriptor` into `NonfungiblePositionLibrary` and make `tokenURI` function a library function (commit [1f70e13](https://github.com/ScopeLift/uniswap-v3-periphery-ovm/commit/1f70e13c7842a4e249c499d2e4c26545138fcfc7))
        - No longer instantiates `PoolKey` struct in memory and uses storage references instead (commit [8a4f929](https://github.com/ScopeLift/uniswap-v3-periphery-ovm/commit/8a4f92905783492a80226abec965cfb9af0196eb))
        - Switch `isAuthorizedForToken` modifier to a private function (commit [1e284f3](https://github.com/ScopeLift/uniswap-v3-periphery-ovm/commit/1e284f383b6fdf452152d3dad12b3ffeece782de))
- Introduce a new CI job for the OVM unit tests
- Modifies the hardhat config the same was as described in the [V3 Core PR](https://github.com/Uniswap/uniswap-v3-core/pull/491)
- Uses `hardhat-dependency-compiler` plugin to compile the V3 Core contracts for the OVM since those artifacts are required for testing
- Remove all `Promise.all()` usage when sending transactions to ensure nonces are properly updated
- Adds a `scripts/run-optimism.sh` script which will clone the Optimism monorepo, build it, and bring up the network
- Because contracts now require library linking (to reduce contract size), and there is a hardcoded constant bytecode hash, the following changes were made:
    - Adds a `_Setup.spec.ts` test file which must always be run first. This ensures library addresses (which are dependent on account nonce) are always the same, which ensures the bytecode hash is the same on each test run
    - Adds `set-hash:evm` and `set-hash:ovm` scripts to update the hardcoded bytecode hash in `PoolAddress.sol`. These are run automatically before running tests
    - Modify `v3CoreFactoryFixture` so it's only run once per test suite
    - Patch waffle's `deployMockContract` so it supports mock OVM contracts
    - New instructions in `README.md` on how to run the EVM and OVM tests to accomodate this

## Status

All tests pass both in the EVM and in the OVM, with the exception of `V3Migrator`, which is not required to be deployed on the OVM. A few tests are skipped:

- `owned by eoa` permit tests are skipped on the EVM since only EIP-1271 contract wallets are supported
- `displays ETH as token symbol for WETH token` are skipped on the EVM since it fails due to the hardcoded WETH9 address in `NonfungiblePositionLibrary`
- `V3Migrator` tests are skipped on the OVM. They will fail on the OVM since it tries to deploy EVM bytecode. (If V3Migrator is needed, the current blocker is adding support to the `@eth-optimism/hardhat-ovm` plugin to support compiling with multiple OVM solc versions)

Due to the lack of snapshots and running against a real node, tests take much longer, ~40mins.

h/t @elenadimitrova @k-ho @ben-chain @smartcontracts @gakonst for their help


NoahZinsmeister:
merging this because of permissioning issues, it still needs some work

PR #150, opened 2021-06-21
fix CI compile issue: Contract code size exceeds 24576 bytes

ewilz:
all better: 
Screen Shot 2021-06-21 at 1 36 39 PM


moodysalem:
you can also reference the interface instead of the contract itself, and then solidity won't compile it

ewilz:
yah that sounds better

moodysalem (approved):

Issue #144, opened 2021-06-11
Create library to compute the token0/token1 value of a NFT


moodysalem:
Could have a function that returns the value of the principal and separately one that returns the value of the fees, and maybe one that returns both

PR #152, opened 2021-06-23
Position Value Library

ewilz:
fixes https://github.com/Uniswap/uniswap-v3-periphery/issues/144

### DIFF
--- a/contracts/libraries/PositionValue.sol
+++ b/contracts/libraries/PositionValue.sol
@@ -16,8 +16,47 @@ library PositionValue {
         view
         returns (uint256 amount0, uint256 amount1)
     {
-        (amount0, amount1) = principal(nft, tokenId);
-        (uint256 amount0Fees, uint256 amount1Fees) = fees(nft, tokenId);
+        (
+            ,
+            ,
+            address token0,
+            address token1,
+            uint24 fee,
+            int24 tickLower,
+            int24 tickUpper,
+            uint128 liquidity,
+            uint256 positionFeeGrowthInside0LastX128,
+            uint256 positionFeeGrowthInside1LastX128,
+            uint256 tokensOwed0,
+            uint256 tokensOwed1
+        ) = nft.positions(tokenId);
+
+        (uint160 sqrtRatioX96, , , , , , ) =
+            IUniswapV3Pool(
+                PoolAddress.computeAddress(
+                    nft.factory(),
+                    PoolAddress.PoolKey({token0: token0, token1: token1, fee: fee})
+                )
+            )
+                .slot0();
+
+        (amount0, amount1) = _fees(
+            nft,
+            PositionParams(
+                token0,
+                token1,
+                fee,
+                tickLower,
+                tickUpper,
+                liquidity,
+                positionFeeGrowthInside0LastX128,
+                positionFeeGrowthInside1LastX128,
+                tokensOwed0,
+                tokensOwed1
+            )
+        );
+        (uint256 amount0Principal, uint256 amount1Principal) =
+            _principal(PrincipalParams(sqrtRatioX96, tickLower, tickUpper, liquidity));
         amount0 += amount0Fees;
         amount1 += amount1Fees;
     }
@@ -39,12 +78,23 @@ library PositionValue {
             )
                 .slot0();
 
+        return _principal(PrincipalParams(sqrtRatioX96, tickLower, tickUpper, liquidity));
+    }
+
+    struct PrincipalParams {
+        uint160 sqrtRatioX96;
+        int24 tickLower;
+        int24 tickUpper;
+        uint128 liquidity;
+    }
+
+    function _principal(PrincipalParams memory params) internal view returns (uint256 amount0, uint256 amount1) {
         return
             LiquidityAmounts.getAmountsForLiquidity(
-                sqrtRatioX96,
-                TickMath.getSqrtRatioAtTick(tickLower),
-                TickMath.getSqrtRatioAtTick(tickUpper),
-                liquidity
+                params.sqrtRatioX96,
+                TickMath.getSqrtRatioAtTick(params.tickLower),
+                TickMath.getSqrtRatioAtTick(params.tickUpper),
+                params.liquidity
             );
     }
 
@@ -99,10 +149,11 @@ library PositionValue {
         uint256 tokensOwed1;
     }
 
-    function _fees(
-        INonfungiblePositionManager nft,
-        PositionParams memory positionParams
-    ) private view returns (uint256 amount0, uint256 amount1) {
+    function _fees(INonfungiblePositionManager nft, PositionParams memory positionParams)
+        private
+        view
+        returns (uint256 amount0, uint256 amount1)
+    {
         (uint256 poolFeeGrowthInside0LastX128, uint256 poolFeeGrowthInside1LastX128) =
             _getFeeGrowthInside(
                 nft,
@@ -113,17 +164,21 @@ library PositionValue {
                 positionParams.tickUpper
             );
 
-        amount0 = FullMath.mulDiv(
-            poolFeeGrowthInside0LastX128 - positionParams.positionFeeGrowthInside0LastX128,
-            positionParams.liquidity,
-            FixedPoint128.Q128
-        ) + positionParams.tokensOwed0;
+        amount0 =
+            FullMath.mulDiv(
+                poolFeeGrowthInside0LastX128 - positionParams.positionFeeGrowthInside0LastX128,
+                positionParams.liquidity,
+                FixedPoint128.Q128
+            ) +
+            positionParams.tokensOwed0;
 
-        amount1 = FullMath.mulDiv(
-            poolFeeGrowthInside1LastX128 - positionParams.positionFeeGrowthInside1LastX128,
-            positionParams.liquidity,
-            FixedPoint128.Q128
-        ) + positionParams.tokensOwed1;
+        amount1 =
+            FullMath.mulDiv(
+                poolFeeGrowthInside1LastX128 - positionParams.positionFeeGrowthInside1LastX128,
+                positionParams.liquidity,
+                FixedPoint128.Q128
+            ) +
+            positionParams.tokensOwed1;
     }
 
     function _getFeeGrowthInside(
--- a/contracts/libraries/PositionValue.sol
+++ b/contracts/libraries/PositionValue.sol
@@ -15,8 +15,8 @@ library PositionValue {
     /// @notice returns the total token value including the principal and the fees owed
     /// @param positionManager The Uniswap V3 NonfungiblePositionManager
     /// @param tokenId The tokenId of the token for which to get the total value
-    /// @returns amount0 The total amount of token0 including principal and fees
-    /// @returns amount1 The total amount of token1 including principal and fees
+    /// @return amount0 The total amount of token0 including principal and fees
+    /// @return amount1 The total amount of token1 including principal and fees
     function total(INonfungiblePositionManager positionManager, uint256 tokenId)
         internal
         view
@@ -78,8 +78,8 @@ library PositionValue {
     /// that the position is burned
     /// @param positionManager The Uniswap V3 NonfungiblePositionManager
     /// @param tokenId The tokenId of the token for which to get the total principal owed
-    /// @returns amount0 The principal amount of token0
-    /// @returns amount1 The principal amount of token1
+    /// @return amount0 The principal amount of token0
+    /// @return amount1 The principal amount of token1
     function principal(INonfungiblePositionManager positionManager, uint256 tokenId)
         internal
         view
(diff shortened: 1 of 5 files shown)


moodysalem:
we actually have this library already that does something very similar
https://github.com/Uniswap/uniswap-v3-periphery/blob/0e8ffedb28909712e76b9c4a94669ca7cfc0e3e7/contracts/libraries/LiquidityAmounts.sol#L120-L136

what we want is to input a token id, and get the position value in token0/token1 with minimal number of calls/sloads

ewilz:
> we actually have this library already that does something very similar
> https://github.com/Uniswap/uniswap-v3-periphery/blob/0e8ffedb28909712e76b9c4a94669ca7cfc0e3e7/contracts/libraries/LiquidityAmounts.sol#L120-L136
> 
> what we want is to input a token id, and get the position value in token0/token1 with minimal number of calls/sloads

mm...minimizing sloads was tough..lmk if it could be better

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '../interfaces/INonfungiblePositionManager.sol';
+import './LiquidityAmounts.sol';
+import './PoolAddress.sol';
+
+library PositionValue {
+    function principal(
+        address factory,
```
moodysalem, line 12: calling `nft.factory()` shouldn't be much worse gas-wise

moodysalem:
> minimizing sloads was tough

yeah there's not much you can do. ideally we would have something like this so you don't sload anything you don't care about: https://eips.ethereum.org/EIPS/eip-2330

test/PositionValue.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+      // console.log(fees.amount1.toString())
+
+      const pool = new ethers.Contract(
+        computePoolAddress(await nft.factory(), [tokens[0].address, tokens[1].address], FeeAmount.MEDIUM),
+        IUniswapV3PoolABI,
+        wallets[0]
+      )
+
+      // this reverts with "NP"
+      await pool.burn(getMinTick(TICK_SPACINGS[FeeAmount.MEDIUM]), getMaxTick(TICK_SPACINGS[FeeAmount.MEDIUM]), 0, {
```
ewilz, line 216: I need to update the pool position to get up-to-date fee amounts. Any idea why collect (L200) works just fine, but I fall into a revert when I burn 0 liquidity here?? (cc @moodysalem )

moodysalem: position owner is the nft, you can't burn it from a wallet

moodysalem: this is probably the trickiest part of computing fees, unfortunately

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,88 @@
...
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/FixedPoint128.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '../interfaces/INonfungiblePositionManager.sol';
+import './LiquidityAmounts.sol';
+import './PoolAddress.sol';
+import './PositionKey.sol';
+
+library PositionValue {
```
moodysalem, line 12: natspec?
also what about a total method?

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,88 @@
...
+    function principal(INonfungiblePositionManager nft, uint256 tokenId)
+        internal
+        view
+        returns (uint256 amount0, uint256 amount1)
+    {
+        (, , address token0, address token1, uint24 fee, int24 tickLower, int24 tickUpper, uint128 liquidity, , , , ) =
+            nft.positions(tokenId);
```
moodysalem, line 19: what about tokens owed to the position but not collected returned by this method? include in fees or principal? i think fees because they are not subject to IL

moodysalem (changes requested):

bitlovk354 (commented):
ok

test/PositionValue.spec.ts
moodysalem: just to elaborate, since i came back and forgot what i was talking about

pool.burn is being called from a wallet, which has no position at the min/max tick, so pool#burn throws NP for 'no position'
https://github.com/Uniswap/uniswap-v3-core/blob/f03155670ec1667406b83a539e23dcccf32a03bc/contracts/libraries/Position.sol#L54

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,88 @@
...
+        );
+
+        amount1 = FullMath.mulDiv(
+            poolFeeGrowthInside1LastX128 - positionFeeGrowthInside1LastX128,
+            liquidity,
+            FixedPoint128.Q128
+        );
+    }
+
+    function getFeeGrowthInside(
```
moodysalem, line 75: you want the current fee growth inside for the tick range, not the last time the nft's position was poked (e.g. via an nft burn or mint at the same tick range)

i'm pretty sure this requires computing this

https://github.com/Uniswap/uniswap-v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/contracts/libraries/Tick.sol#L60-L95

moodysalem: you'll want to do it more like this so you can avoid reading the fee growth globals if you're above/below the tick range
https://github.com/Uniswap/uniswap-v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/contracts/UniswapV3Pool.sol#L202-L233

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,88 @@
...
+library PositionValue {
+    function principal(INonfungiblePositionManager nft, uint256 tokenId)
+        internal
+        view
+        returns (uint256 amount0, uint256 amount1)
+    {
+        (, , address token0, address token1, uint24 fee, int24 tickLower, int24 tickUpper, uint128 liquidity, , , , ) =
+            nft.positions(tokenId);
+
+        (uint160 sqrtRatioX96, , , , , , ) =
```
moodysalem, line 21: need to allow passing a price into this function, maybe a totally separate function that just takes the price? this price is subject to manipulation

moodysalem (changes requested):

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,167 @@
...
+import '@uniswap/v3-core/contracts/libraries/FixedPoint128.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@uniswap/v3-core/contracts/libraries/Tick.sol';
+import '../interfaces/INonfungiblePositionManager.sol';
+import './LiquidityAmounts.sol';
+import './PoolAddress.sol';
+import './PositionKey.sol';
+
+library PositionValue {
+    function total(INonfungiblePositionManager nft, uint256 tokenId)
```
ewilz, line 14: this is not in its final form btw

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,226 @@
...
+    function _getFeeGrowthInside(
+        INonfungiblePositionManager positionManager,
+        address token0,
+        address token1,
+        uint24 fee,
+        int24 tickLower,
+        int24 tickUpper
+    ) private view returns (uint256 feeGrowthInside0X128, uint256 feeGrowthInside1X128) {
+        IUniswapV3Pool pool =
+            IUniswapV3Pool(
```
moodysalem, line 167: seems like we should just be passing around pool in the fee params, rather than token0/token1/fee

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,226 @@
...
+            uint256 feeGrowthGlobal1X128 = pool.feeGrowthGlobal1X128();
+            feeGrowthInside0X128 = feeGrowthGlobal0X128 - lowerFeeGrowthOutside0X128 - upperFeeGrowthOutside0X128;
+            feeGrowthInside1X128 = feeGrowthGlobal1X128 - lowerFeeGrowthOutside1X128 - upperFeeGrowthOutside1X128;
+        } else {
+            feeGrowthInside0X128 = upperFeeGrowthOutside0X128 - lowerFeeGrowthOutside0X128;
+            feeGrowthInside1X128 = upperFeeGrowthOutside1X128 - lowerFeeGrowthOutside1X128;
+        }
+    }
+
+    function _constructParams(
```
moodysalem, line 192: somewhat prefer just taking a gas hit and making this more readable since it's for public consumption, by just having fees and principal be public methods and having total just do `fees(nft, tokenId) + principal(nft,tokenId,sqrtRatioX96)`

if you can flatten the code out a bit as a result and remove the passing around of large structs, it might be a negligible difference

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,226 @@
...
+        int24 tickLower;
+        int24 tickUpper;
+        uint128 liquidity;
+    }
+
+    /// @notice Calculates the principal (currently acting as liquidity) owed to the token owner in the event
+    /// that the position is burned
+    /// @param positionManager The Uniswap V3 NonfungiblePositionManager
+    /// @param tokenId The tokenId of the token for which to get the total principal owed
+    /// @param sqrtRatioX96 The square root price X96 for which to calculate the principal amounts
```
moodysalem, line 47: this is going to be hard to compute without additional guidance, we should create another library for going from some ratio (e.g. `1e6:1e18`, or `1 with 6 decimals : 1 with 18 decimals`) to sqrt ratio x96 after this pr

contracts/libraries/PositionValue.sol
ewilz: I already had them defined though. You think I should sload them again from the pool?

moodysalem: i mean that we only use token0/token1/fee to compute the pool, but we are passing around 3 variables instead of the 1 pool address, so you can compute pool in the caller and pass it in to _getFeeGrowthInside instead of all 4 of 
````
        INonfungiblePositionManager positionManager,
        address token0,
        address token1,
        uint24 fee,
```

contracts/libraries/PositionValue.sol
```diff
@@ -23,23 +24,13 @@ library PositionValue {
         uint256 tokenId,
         uint160 sqrtRatioX96
     ) internal view returns (uint256 amount0, uint256 amount1) {
-        (PrincipalParams memory principalParams, FeeParams memory feeParams) =
-            _constructParams(positionManager, tokenId, sqrtRatioX96);
+        (amount0, amount1) = principal(positionManager, tokenId, sqrtRatioX96);
```
moodysalem, line 27: readability nit: principal0, principal1 and fees0, fees1, then just `return (principal0 + fees0, principal + fees1)`

contracts/libraries/PositionValue.sol
ewilz: ohhhhh

ewilz:
gas pre-readability refactory
```
exports[`PositionValue #fees when price is above the position range gas 1`] = `47680`;

exports[`PositionValue #fees when price is below the position range gas 1`] = `47648`;

exports[`PositionValue #fees when price is within the position range gas 1`] = `53238`;

exports[`PositionValue #principal gas 1`] = `23209`;

exports[`PositionValue #total gas 1`] = `56939`;
```

gas post-refactor 
```
exports[`PositionValue #fees when price is above the position range gas 1`] = `47663`;

exports[`PositionValue #fees when price is below the position range gas 1`] = `47631`;

exports[`PositionValue #fees when price is within the position range gas 1`] = `53221`;

exports[`PositionValue #principal gas 1`] = `23001`;

exports[`PositionValue #total gas 1`] = `59782`;
```

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,169 @@
...
+            FullMath.mulDiv(
+                poolFeeGrowthInside1LastX128 - feeParams.positionFeeGrowthInside1LastX128,
+                feeParams.liquidity,
+                FixedPoint128.Q128
+            ) +
+            feeParams.tokensOwed1;
+    }
+
+    function _getFeeGrowthInside(
+        INonfungiblePositionManager positionManager,
```
moodysalem, line 147: no longer used

contracts/libraries/PositionValue.sol
```diff
@@ -0,0 +1,169 @@
...
+                    liquidity: liquidity,
+                    positionFeeGrowthInside0LastX128: positionFeeGrowthInside0LastX128,
+                    positionFeeGrowthInside1LastX128: positionFeeGrowthInside1LastX128,
+                    tokensOwed0: tokensOwed0,
+                    tokensOwed1: tokensOwed1
+                })
+            );
+    }
+
+    function _fees(INonfungiblePositionManager positionManager, FeeParams memory feeParams)
```
moodysalem, line 111: i wish we could get rid of this private method but i guess you got stack too deep from it

ewilz: yes. Pulling 10 parameters from `positions()` makes it pretty difficult to dance around a stack too deep

PR #159, opened 2021-06-30
WIP: Swap and Add

ewilz:


contracts/libraries/PoolTicksLibrary.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.6.0;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/BitMath.sol';
+
+library PoolTicksLibrary {
```
ewilz, line 7: this library has the exact logic I need. However it performs changes on storage, so I switched it to behave on behalf of an `IUniswapV3Pool`

ewilz:
closing for now

PR #169, opened 2021-07-15
first pass liquidity samples

snreynolds:


contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,249 @@
...
+                recipient: address(this),
+                tickLower: deposits[tokenId].tickLower,
+                tickUpper: deposits[tokenId].tickUpper,
+                amount0Desired: amountAdd0,
+                amount1Desired: amountAdd1,
+                amount0Min: 0,
+                amount1Min: 0
+            });
+
+        (liquidity, amount0, amount1, pool) = addLiquidity(params);
```
ewilz, line 207: take a look at where the nftManager uses`addLiquidity`. It does a lot of things to update the position. I think you're better off calling the nft manager directly for this: `nftManager.increaseLiquidity()`

in this case you'd be adding liquidity on behalf of this contract, not behalf of an nft.

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,249 @@
...
+    /// @return amount1 The amount returned back in token1
+    function decreaseLiquidityInHalf(uint256 tokenId) external returns (uint256 amount0, uint256 amount1) {
+        // caller must be the owner of the NFT
+        require(msg.sender == deposits[tokenId].owner, 'Not the owner');
+        // get liquidity data for tokenId
+        uint128 liquidity = deposits[tokenId].liquidity;
+        uint128 halfLiquidity = liquidity / 2;
+
+        // amount0Min and amount1Min are price slippage checks
+        // if the amount received after burning is not greater than these minimums, transaction will fail
```
ewilz, line 159: it actually might be fun to add a slippage check in the contract. say for instance decreaseLiquidityForStablePair. The liquidity amount could be a param or you could just do it in half again. This would be tricky though esp since some tokens like USDC uses 6 decimals -__- 

Ok this is actually prettty annoyingly tricky, and it might even merit its own example contract someday.

You'd have to find out token decimals from each token: `1e6` is 1USDC and `1e18` is 1DAI.
Then the price of 1 would be multiplied or divided by the difference in decimals depending on which token is token0.
Then you could use different libraries like: `LiquidityAmounts.getAmount0ForLiquidity()`and have a minimum that amount to get back which would signify how much of the price of 1 the current price is.

(not a blocker)

ewilz (commented):
Looking good! A few comments, some changes requested, others just thoughts.

I'll look a little deeper later bc I might have missed stuff!

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,249 @@
...
+        address token1 = deposits[tokenId].token1;
+        // send collected fees to owner
+        TransferHelper.safeTransfer(token0, owner, amount0);
+        TransferHelper.safeTransfer(token1, owner, amount1);
+    }
+
+    /// @notice Transfers the NFT to this contract
+    /// @dev The transfer triggers `onERC721Received`
+    /// @param tokenId The id of the erc721
+    function despositNFT(uint256 tokenId) external {
```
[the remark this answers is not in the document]
snreynolds: Makes sense! Originally I had the logic of building the mapping of token -> owner in here but realized that should happen in the `onERC721Received` function. So now I see that this function is pretty useless since an external contract can just call transfer to this contract, triggering `onERC721Received`

contracts/examples/LiquidityExamples.sol
snreynolds: I love this - yes we should definitely include this. Going to push the changes above for now so we can get the MVP guide up and will work on this as a next addition. Might need to run some questions by you so will leave this as unresolved.

ewilz: I can resolve for now. This feature is probably better for another time??

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
...
+    function _createDeposit(address owner, uint256 tokenId) internal {
+        (, , address token0, address token1, , , , uint128 liquidity, , , , ) =
+            nonfungiblePositionManager.positions(tokenId);
+
+        // set the owner and data for position
+        // operator is msg.sender
+        deposits[tokenId] = Deposit({owner: owner, liquidity: liquidity, token0: token0, token1: token1});
+    }
+
+    /// @notice Calls the mint function defined in perphery, mints the same amount of each token. For this example we are providing 1000 DAI and 1000 USDC in liquidity
```
ewilz, line 63: should probably just multi-line this

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
...
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol';
+import '../libraries/TransferHelper.sol';
+import '../interfaces/ISwapRouter.sol';
+import '../interfaces/INonfungiblePositionManager.sol';
+import '../base/LiquidityManagement.sol';
+
+contract LiquidityExamples is IERC721Receiver, LiquidityManagement {
```
ewilz, line 13: does `LiquidityManagement` still need to be extended?

contracts/examples/LiquidityExamples.sol
snreynolds: yes. I'll work on this starting next week :)

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity =0.7.6;
+pragma abicoder v2;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol';
+import '../libraries/TransferHelper.sol';
+import '../interfaces/ISwapRouter.sol';
```
ewilz, line 9: leftover, not using ISwapRouter??

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
...
+            uint256 amount1
+        )
+    {
+        // For this example, we will provide equal amounts of liquidity in both assets.
+        // Providing liquidity in both assets means liquidity will be earning fees and is considered in-range.
+        uint256 amount0ToMint = 1000;
+        uint256 amount1ToMint = 1000;
+
+        // Approve the position manager
+        TransferHelper.safeApprove(DAI, address(nonfungiblePositionManager), amount0ToMint);
```
ewilz, line 83: when did the tokens make their way into the contract here?

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
...
+            TransferHelper.safeTransfer(USDC, msg.sender, refund1);
+        }
+    }
+
+    /// @notice Collects the fees associated with provided liquidity
+    /// @dev The contract must hold the erc721 token before it can collect fees
+    /// @param tokenId The id of the erc721 token
+    /// @return amount0 The amount of fees collected in token0
+    /// @return amount1 The amount of fees collected in token1
+    function collectAllFees(uint256 tokenId) external returns (uint256 amount0, uint256 amount1) {
```
ewilz, line 126: Might make more sense to collect the fees from an existing deposit, rather than create a new deposit while collecting fees? Seems a bit backwards in a way...

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,234 @@
...
+            INonfungiblePositionManager.IncreaseLiquidityParams({
+                tokenId: tokenId,
+                amount0Desired: amountAdd0,
+                amount1Desired: amountAdd1,
+                amount0Min: 0,
+                amount1Min: 0,
+                deadline: block.timestamp
+            });
+
+        (liquidity, amount0, amount1) = nonfungiblePositionManager.increaseLiquidity(params);
```
ewilz, line 202: as @Yorkemartin  noticed....need to approve erc20 tokens

terapi09 (approved):

coleymarie1 (approved):

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,256 @@
...
+>>>>>>> 500f18e847adf00ee69f1fff7ea9566e98aa1476
```
ewilz, line 224: woops. don't forget to try `yarn compile` !!

contracts/examples/LiquidityExamples.sol
ewilz: you can always get a second opinion on the idea too...other things may be a bigger priority

hamdan33-com (approved):

hamdan33-com (approved):
https://github.com/moodysalem/java-phantomjs-wrapper/issues/13#issue-919632501

hamdan33-com (commented):
Thx u

hamdan33-com (approved):

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /// @notice Transfers funds to owner of NFT
+    /// @param tokenId The id of the erc721
+    /// @param amount0 The amount of token0
+    /// @param amount1 The amount of token1
+    function _sendToOwner(
```
hamdan33-com, line 231: clear

### 

- 

- [ ]

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,238 @@
...
+        TransferHelper.safeApprove(DAI, address(nonfungiblePositionManager), amount0ToMint);
+        TransferHelper.safeApprove(USDC, address(nonfungiblePositionManager), amount1ToMint);
+
+        INonfungiblePositionManager.MintParams memory params =
+            INonfungiblePositionManager.MintParams({
+                token0: DAI,
+                token1: USDC,
+                fee: poolFee,
+                tickLower: TickMath.MIN_TICK,
+                tickUpper: TickMath.MAX_TICK,
```
NoahZinsmeister, line 85: i believe these values won't work for all tick spacings? maybe just worth noting in a comment

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,238 @@
...
+import '../interfaces/INonfungiblePositionManager.sol';
+import '../base/LiquidityManagement.sol';
+
+contract LiquidityExamples is IERC721Receiver {
+    address public constant DAI = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+    address public constant USDC = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
+
+    uint24 public constant poolFee = 3000;
+
+    INonfungiblePositionManager public immutable nonfungiblePositionManager;
```
NoahZinsmeister: bit odd for this to be the only immutable, i'd just hardcode all of DAI/USDC/NFT manager to be dummy addresses

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,238 @@
...
+            INonfungiblePositionManager.MintParams({
+                token0: DAI,
+                token1: USDC,
+                fee: poolFee,
+                tickLower: TickMath.MIN_TICK,
+                tickUpper: TickMath.MAX_TICK,
+                amount0Desired: amount0ToMint,
+                amount1Desired: amount1ToMint,
+                amount0Min: 0,
+                amount1Min: 0,
```
NoahZinsmeister, line 89: should comment that this is unsafe

contracts/examples/LiquidityExamples.sol
```diff
@@ -0,0 +1,238 @@
...
+        INonfungiblePositionManager.IncreaseLiquidityParams memory params =
+            INonfungiblePositionManager.IncreaseLiquidityParams({
+                tokenId: tokenId,
+                amount0Desired: amountAdd0,
+                amount1Desired: amountAdd1,
+                amount0Min: 0,
+                amount1Min: 0,
+                deadline: block.timestamp
+            });
+
```
NoahZinsmeister, line 197: need to unapprove after this call like in mintNewPosition

contracts/examples/LiquidityExamples.sol
```diff
@@ -33,25 +29,22 @@ contract LiquidityExamples is IERC721Receiver {
...
+    // Note that the operator is recorded as the owner of the deposited NFT
     function onERC721Received(
         address operator,
         address,
         uint256 tokenId,
         bytes calldata
     ) external override returns (bytes4) {
-        // get position information
-
+        require(msg.sender == address(nonfungiblePositionManager), 'not a univ3 nft');
```
NoahZinsmeister, line 39: important safety check

contracts/examples/LiquidityExamples.sol
```diff
@@ -230,9 +218,9 @@ contract LiquidityExamples is IERC721Receiver {
     function retrieveNFT(uint256 tokenId) external {
         // must be the owner of the NFT
         require(msg.sender == deposits[tokenId].owner, 'Not the owner');
+        // remove information related to tokenId
+        delete deposits[tokenId];
```
NoahZinsmeister, line 226: very important, always have to follow the checks/effects/interactions pattern

NoahZinsmeister (commented):
left some comments and also pushed a commit with important safety changes. LGTM after feedback is addressed, but as discussed offline let's close this PR and have this code live in the docs

Thefy21 (approved):

Thefy21 (approved):

PR #171, opened 2021-07-19
Create sdkPoolExamples.ts

Yorkemartin:
Never got this properly reviewed in the previous cycle - the goal is a first-principles example of creating a pool instance, mostly to contextualize the v3 sdk VS ethers.js

Couldn't figure out how to promise.all the correct way in `getPoolState`, could use some advice there.

### DIFF
--- a/contracts/examples/sdkPoolExamples.ts
+++ b/contracts/examples/sdkPoolExamples.ts
@@ -1,0 +1,83 @@
+
+import { ethers } from "ethers";
+import { Pool } from "@uniswap/v3-sdk";
+import { Token } from "@uniswap/sdk-core";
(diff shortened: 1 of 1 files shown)


moodysalem:
this shouldn't be in periphery, periphery doesn't depend on v3-sdk (it's the other way around) and this seems to be an example of creating a v3 sdk Pool instance rather than creating a pool on ethereum

Yorkemartin:
@moodysalem Noah / Sara / I previously discussed the best route for getting reviews for non-production code examples that will be used for docs. We decided that draft reviews in the v3 periphery repo was the best way to allow review, the pr's will be closed and not merged when the code is ready for the docs. 

But yes - this is supposed to be a very basic example of creating a pool instance using the v3 sdk

moodysalem:
what about the docs repo where the code samples will live? why the periphery repo?

PR #180, opened 2021-08-03
On-chain Swap + Add

ewilz:
- [ ] improve `swapToNextInitializedTick` to more closely match real swaps ( will try SwapMath lib from core)
- [x] `nextInitializedTickWithinOneWord` to work going backwards
- [ ] "gas golfing"
- [ ] implement binary search

### DIFF
--- a/contracts/libraries/PoolTicksLibrary.sol
+++ b/contracts/libraries/PoolTicksLibrary.sol
@@ -5,14 +5,14 @@ import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
 import '@uniswap/v3-core/contracts/libraries/BitMath.sol';
 
 library PoolTicksLibrary {
-		/// @notice Computes the position in the mapping where the initialized bit for a tick lives
-		/// @param tick The tick for which to compute the position
-		/// @return wordPos The key in the mapping containing the word in which the bit is stored
-		/// @return bitPos The bit position in the word where the flag is stored
-		function position(int24 tick) private pure returns (int16 wordPos, uint8 bitPos) {
-				wordPos = int16(tick >> 8);
-				bitPos = uint8(tick % 256);
-		}
+    /// @notice Computes the position in the mapping where the initialized bit for a tick lives
+    /// @param tick The tick for which to compute the position
+    /// @return wordPos The key in the mapping containing the word in which the bit is stored
+    /// @return bitPos The bit position in the word where the flag is stored
+    function position(int24 tick) private pure returns (int16 wordPos, uint8 bitPos) {
+        wordPos = int16(tick >> 8);
+        bitPos = uint8(tick % 256);
+    }
 
     /// @notice Returns the next initialized tick contained in the same word (or adjacent word) as the tick that is either
     /// to the left (less than or equal to) or right (greater than) of the given tick
@@ -22,7 +22,7 @@ library PoolTicksLibrary {
     /// @param lte Whether to search for the next initialized tick to the left (less than or equal to the starting tick)
     /// @return next The next initialized or uninitialized tick up to 256 ticks away from the current tick
     /// @return initialized Whether the next tick is initialized, as the function only searches within up to 256 ticks
-		///
+    ///
     function nextInitializedTickWithinOneWord(
         IUniswapV3Pool pool,
         int24 tick,
--- a/contracts/libraries/PoolTicksLibrary.sol
+++ b/contracts/libraries/PoolTicksLibrary.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.6.0;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/BitMath.sol';
+
+library PoolTicksLibrary {
+		/// @notice Computes the position in the mapping where the initialized bit for a tick lives
+		/// @param tick The tick for which to compute the position
+		/// @return wordPos The key in the mapping containing the word in which the bit is stored
+		/// @return bitPos The bit position in the word where the flag is stored
+		function position(int24 tick) private pure returns (int16 wordPos, uint8 bitPos) {
+				wordPos = int16(tick >> 8);
+				bitPos = uint8(tick % 256);
+		}
+
+    /// @notice Returns the next initialized tick contained in the same word (or adjacent word) as the tick that is either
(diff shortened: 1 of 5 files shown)


test/__snapshots__/SwapToRatioTest.spec.ts.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`SwapToRatio #getPostSwapPrice  when initial deposit has excess of token1 and swap pushes the price beyond multiple initialized ticks gas 1`] = `50782`;
```
ewilz, line 3: not the greatest...

contracts/libraries/PoolTicksLibrary.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.6.0;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/BitMath.sol';
+
+library PoolTicksLibrary {
```
ewilz, line 7: copied this from core and takes pool as an arg instead of a storage mapping.

test/SwapToRatioTest.spec.ts
```diff
@@ -0,0 +1,662 @@
...
+            tokenOut: token1.address,
+            fee: FeeAmount.MEDIUM,
+            recipient: wallets[0].address,
+            deadline: 1,
+            amountIn: amountToSwap.abs(),
+            amountOutMinimum: 0,
+            sqrtPriceLimitX96: expandTo18Decimals(100_000),
+          })
+
+          expect((await pool.slot0()).sqrtPriceX96).to.eq(sqrtRatioX96)
```
ewilz, line 593:
```
AssertionError: Expected "78991555868274566570269357078" to be equal 79228162514264337593543950336
```

would work backwards from swapAmounts first to see if this improves in the process

contracts/libraries/SwapToRatio.sol
```diff
@@ -0,0 +1,189 @@
...
+    {
+        if (zeroForOne) {
+            if (sqrtRatioX96Target < positionParams.sqrtRatioX96Lower) {
+                return (false, 0, 0);
+            }
+        } else {
+            if (sqrtRatioX96Target > positionParams.sqrtRatioX96Upper) {
+                return (false, 0, 0);
+            }
+        }
```
ewilz, line 57: would not need to perform for rebalancing out of range position

ewilz: haven't really addressed the case for an our of range position. Will likely get refactored

contracts/libraries/SwapToRatio.sol
```diff
@@ -0,0 +1,189 @@
...
+        // return calculateConstantLiquidityPostSwapSqrtPrice(poolParams, positionParams);
+    }
+
+    function isZeroForOne(
+        uint256 amount0,
+        uint256 amount1,
+        uint160 sqrtRatioX96,
+        uint160 sqrtRatioX96Lower,
+        uint160 sqrtRatioX96Upper
+    ) internal view returns (bool) {
```
NoahZinsmeister, line 160: for precision purposes, we might consider doing some of:

1. transform this from a/b > c/d to a > (c*b)/d, where the RHS uses mulDiv
2. use a much larger liquidity value (as large as possible while avoiding overflow)
3. think through whether we should round c and d in different directions, to bias the result in one direction or the other (would need to think through this one)

some/all of these might require bounding amounts to uint128 or similar

contracts/libraries/SwapToRatio.sol
```diff
@@ -0,0 +1,189 @@
...
+        uint256 amount0Next;
+        uint256 amount1Next;
+        int24 nextInitializedTick;
+        uint160 sqrtRatioNextX96;
+
+        while (crossTickBoundary) {
+            // returns the next initialized tick or the last tick within one word of the current tick.
+            // We'll renew calculation at least on a per word basis for better rounding
+            (nextInitializedTick, ) = pool.nextInitializedTickWithinOneWord(tick, tickSpacing, zeroForOne);
+            if (zeroForOne) nextInitializedTick -= 1;
```
NoahZinsmeister, line 125: sorry, why do we need this? isn't it handled by `nextInitializedTickWithinOneWord`?

contracts/libraries/SwapToRatio.sol
```diff
@@ -0,0 +1,189 @@
...
+        bool crossTickBoundary = true;
+        uint256 amount0Next;
+        uint256 amount1Next;
+        int24 nextInitializedTick;
+        uint160 sqrtRatioNextX96;
+
+        while (crossTickBoundary) {
+            // returns the next initialized tick or the last tick within one word of the current tick.
+            // We'll renew calculation at least on a per word basis for better rounding
+            (nextInitializedTick, ) = pool.nextInitializedTickWithinOneWord(tick, tickSpacing, zeroForOne);
```
NoahZinsmeister, line 124: ok, this might be really annoying, but in theory every tick of a word could be initialized, and iiuc we would be making repeat calls every time to fetch the whole word in that case? need an idempotent fetch with caching 😬😬😬

contracts/libraries/SwapToRatio.sol
```diff
@@ -0,0 +1,189 @@
...
+            );
+        uint256 validDeposit1 =
+            SqrtPriceMath.getAmount1Delta(
+                sqrtRatioX96Target,
+                positionParams.sqrtRatioX96Lower,
+                poolParams.liquidity,
+                false
+            );
+
+        if (zeroForOne) {
```
NoahZinsmeister, line 89: hmm, i wonder what the gas overhead would be if we just used the quoter here? iuuc this code is basically mocking out the results of a swap?

contracts/libraries/SwapToRatio.sol
ewilz: from what I've seen, when I'm working backwards, it just returns to me the current tick, then infinite loop. Though I've only tested with tick 0.

contracts/libraries/SwapToRatio.sol
NoahZinsmeister: scratch that, for the sake of efficiency we have to manually implement the logic if we only want to run tick transition logic once per function call, the quoter runs it every time

NoahZinsmeister:
closing PR for house-keeping, but will keep the branch around for reference if we ever need it in the future

PR #188, opened 2021-08-12
move deadline out to its own function and add the ability to check the previous block hash

moodysalem:
might want to do breaking changes with major version releases? not sure


NoahZinsmeister:
https://github.com/Uniswap/swap-router-contracts/pull/1

PR #191, opened 2021-08-29
Update deploys.md

Yorkemartin:


NoahZinsmeister:
gonna close this until we feel like quoter v2 is worth sharing around (i think at the moment it might be too specific)

Issue #198, opened 2021-09-07
Pragma of PositionValue library should exclude 0.8.x


moodysalem:
https://github.com/Uniswap/uniswap-v3-periphery/blob/ff3832f52d42a61df4513c1621dfbcfccd8a6d32/contracts/libraries/PositionValue.sol#L2

Our fee calculation code relies on unchecked math that is expected to overflow, need to restrict to `<0.8.0` before publishing and explicitly document it

PR #201, opened 2021-09-09
PositionValue lib: restrict to lower than solc version 8

ewilz:
fixes https://github.com/Uniswap/uniswap-v3-periphery/issues/198

### DIFF
--- a/contracts/libraries/PositionValue.sol
+++ b/contracts/libraries/PositionValue.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
-pragma solidity >=0.7.0 <0.8.0;
+pragma solidity >=0.6.8 <0.8.0;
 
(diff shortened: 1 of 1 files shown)


moodysalem:
what do we need in 0.7.0? would earlier versions work?

ewilz:
> what do we need in 0.7.0? would earlier versions work?

hmm...well I see now our standard is to do from `>=0.5.0` however, I use the `TickMath` library which uses `type(uint256).max` which is new in `Solidity 0.6.8` even though TickMath also claims to compile from `>=0.5.0` But I don't think it will compile lower than 0.6.8

moodysalem (approved):

Bulinlllll:
Tengo [https://github.com/github/docs/commit/9cc87413cb81b8bf94989c8130a4d04c33df1c12?short_path=eca12c0#commitcomment-55251941](position_value_solc)

ewilz:
Changed to 0.6.8

PR #202, opened 2021-09-13
OracleLibrary: Oldest observation helper

### LINKED ISSUE #193: Add function exposing the max secondsAgo to `OracleLibrary`
It may be useful in some cases to know how old the oldest `observation` is, we should consider exposing a function to return the `secondsAgo` associated with the oldest observation (the one that gives the longest possible TWAP)

this should likely be used as a bail-out for ensuring guarantees about minimum TWAP length, as opposed to an actual TWAP period itself

hensha256:
Closes #193

### DIFF
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -58,6 +58,9 @@ library OracleLibrary {
         }
     }
 
+    /// @notice Given a pool, it returns the number of seconds ago of the oldest stored observation
+    /// @param pool Address of Uniswap V3 pool that we want to observe
+    /// @return The number of seconds ago of the oldest observation stored for the pool
     function getOldestObservationSecondsAgo(address pool) internal view returns (uint32) {
         ( , , uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
         require(observationCardinality > 0, 'Pool not initialized');
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -62,12 +62,11 @@ library OracleLibrary {
     /// @param pool Address of Uniswap V3 pool that we want to observe
     /// @return The number of seconds ago of the oldest observation stored for the pool
     function getOldestObservationSecondsAgo(address pool) internal view returns (uint32) {
-        ( , , uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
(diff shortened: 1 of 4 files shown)


contracts/libraries/OracleLibrary.sol
```diff
@@ -57,4 +57,21 @@ library OracleLibrary {
...
+        require(observationCardinality > 0, 'Pool not initialized');
+
+        (uint32 observationTimestamp, , , bool initialized) =
+            IUniswapV3Pool(pool).observations((observationIndex + 1) % observationCardinality);
+
+        if (!initialized) {
+            (observationTimestamp, , , ) = IUniswapV3Pool(pool).observations(0);
+        }
+
+        return uint32(block.timestamp) - observationTimestamp;
```
NoahZinsmeister, line 75: revisiting this after looking at the tests - in core we take in `block.timestamp` as an argument to libraries (e.g. https://github.com/Uniswap/uniswap-v3-core/blob/main/contracts/libraries/Oracle.sol#L147)

doing this here would increase gas costs a bit and probably be a little confusing to callers, so i'm not sure it's worth it (the big benefit is ease of unit testing, both for use and dev users), but something to consider/be aware of.

@moodysalem might have thoughts here

test/OracleLibrary.spec.ts
```diff
@@ -202,4 +202,84 @@ describe('OracleLibrary', () => {
...
+
+      expect(result['secondsAgo']).to.equal(
+        result['currentTimestamp'] - blockTimestamps[(observationIndex + 1) % observationCardinality]
+      )
+    })
+
+    it('fetches from index 0 if the next index is uninitialized', async () => {
+      const blockTimestamps = [1, 2, 0, 0]
+      const initializeds = [true, true, false, false]
+      const observationCardinality = 4
```
NoahZinsmeister, line 254: ditto this should be 2

NoahZinsmeister (changes requested):
nice start! the contract code looks great, left a few testing-related comments. in addition to those, it might be worthwhile to add a test that features overflowed timestamps?

e.g. `blockTimestamps` would be something like `[type(uint32).max - 10, 15, ...]`

contracts/libraries/OracleLibrary.sol
NoahZinsmeister: @moodysalem after chatting with @hensha256 we think that this is the preferred API for callers, so we're going to leave as is and try to get our tests to a spot where they feel clear and thorough - any strong objection to that?

NoahZinsmeister (approved):
LGTM!

PR #204, opened 2021-09-22
OracleLibrary: Block starting tick

### LINKED ISSUE #158: Add gas-optimized function that returns the block-starting tick to `OracleLibrary`
`OracleLibrary` should contain a gas-optimized function (e.g. `consultMostRecent`) which returns the `tick` value as of the start of the current block (end of the last block).

It should first read `slot0`, then fetch `observations[slot0.observationIndex]`. If that observation was made before the current block, we can simply return `slot0.tick`. Otherwise, we need to fetch the prior observation (with the appropriate index modding), and then run the oracle calculation to return the `tick`.

Note: For the purposes of manipulation-resistance, it's almost certainly preferred to use a short-period average over this approach - however, if gas is truly at a premium and manipulation/oracle risk is appropriately constrained, this could make sense.

Note: We could also return the time since the oracle has last updated from this function, which might be another interesting/useful data point in this setting. In the case of the current observation being before the current block, it's time from then to now, otherwise 0. Note that this isn't a perfect indicator, as the oracle can update without the price changing, but it's still highly correlated. In theory, we could actually disambiguate oracle updates from price changes, but in the worst case it would require fetching every other observation, or more realistically doing a full binary search, so probably not worth it.

hensha256:
Closes #158

### DIFF
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -82,22 +82,28 @@ library OracleLibrary {
     /// @return The tick that the pool was in at the start of the current block
     function getBlockStartingTick(address pool) internal view returns (int24) {
         (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
-        require(observationCardinality > 0, 'NI');
+        require(observationCardinality > 0, 'PNI');
 
         // if the latest observation occurred in the past, then no tick-changing trades have happened in this block
         // so the tick in `slot0` is the same as at the beginning of the current block
-        (uint32 observationTimestamp, int56 tickCumulative, , ) = IUniswapV3Pool(pool).observations(observationIndex);
+        (uint32 observationTimestamp, int56 tickCumulative, , bool initialized) = IUniswapV3Pool(pool).observations(observationIndex);
+        require(initialized, 'ONI');
         if (observationTimestamp < block.timestamp) {
             return tick;
         }
 
         // we instead need to fetch the two prior observations and use them to calculate the tick
-        uint256 previousIndex = (observationIndex + observationCardinality - 1) % observationCardinality;
-        (observationTimestamp, tickCumulative, , ) = IUniswapV3Pool(pool).observations(previousIndex);
+        // to do so we require a cardinality of at least 3
+        require(observationCardinality > 2, 'NEO');
 
-        (uint32 prevObservationTimestamp, int56 prevTickCumulative, , ) = IUniswapV3Pool(pool).observations(
-            (previousIndex + observationCardinality - 1) % observationCardinality
+        uint256 prevIndex = (observationIndex + observationCardinality - 1) % observationCardinality;
+        (observationTimestamp, tickCumulative, , initialized) = IUniswapV3Pool(pool).observations(prevIndex);
+
+        (uint32 prevObservationTimestamp, int56 prevTickCumulative, , bool prevInitialized) = IUniswapV3Pool(pool).observations(
+            (prevIndex + observationCardinality - 1) % observationCardinality
         );
+
+        require(initialized && prevInitialized, 'ONI');
 
         // extract the tick from the cumulatives
         return int24((tickCumulative - prevTickCumulative) / (observationTimestamp - prevObservationTimestamp));
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -82,18 +82,17 @@ library OracleLibrary {
     /// @return The tick that the pool was in at the start of the current block
     function getBlockStartingTick(address pool) internal view returns (int24) {
         (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
-        require(observationCardinality > 0, 'PNI');
+
+        // 3 observations are needed to reliably calculate the block starting tick
+        require(observationCardinality > 2, 'NEO');
 
         // If the latest observation occurred in the past, then no tick-changing trades have happened in this block
         // therefore the tick in `slot0` is the same as at the beginning of the current block.
         // We don't need to check if this observation is initialized - it is guaranteed to be.
         (uint32 observationTimestamp, , , ) = IUniswapV3Pool(pool).observations(observationIndex);
-        if (observationTimestamp < block.timestamp) {
+        if (observationTimestamp != block.timestamp) {
             return tick;
         }
-
-        // We need to fetch the two prior observations and use them to calculate the tick.
-        require(observationCardinality > 2, 'NEO');
 
         uint256 prevIndex = (observationIndex + observationCardinality - 1) % observationCardinality;
         int56 tickCumulative;
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -87,8 +87,7 @@ library OracleLibrary {
         // If the latest observation occurred in the past, then no tick-changing trades have happened in this block
         // therefore the tick in `slot0` is the same as at the beginning of the current block.
         // We don't need to check if this observation is initialized - it is guaranteed to be.
-        (uint32 observationTimestamp, , , ) =
-            IUniswapV3Pool(pool).observations(observationIndex);
+        (uint32 observationTimestamp, , , ) = IUniswapV3Pool(pool).observations(observationIndex);
         if (observationTimestamp < block.timestamp) {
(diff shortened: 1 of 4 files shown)


Thefy21 (approved):
Thanks

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,35 @@ library OracleLibrary {
...
+    /// @return The tick that the pool was in at the start of the current block
+    function getBlockStartingTick(address pool) internal view returns (int24) {
+        (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
+        require(observationCardinality > 0, 'PNI');
+
+        // If the latest observation occurred in the past, then no tick-changing trades have happened in this block
+        // therefore the tick in `slot0` is the same as at the beginning of the current block.
+        // We don't need to check if this observation is initialized - it is guaranteed to be.
+        (uint32 observationTimestamp, , , ) = IUniswapV3Pool(pool).observations(observationIndex);
+        if (observationTimestamp < block.timestamp) {
```
moodysalem, line 91: observationTimestamp is a `uint32`, so it can overflow and be less than block.timestamp

recommend `!=` instead

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,35 @@ library OracleLibrary {
...
+            return tick;
+        }
+
+        // We need to fetch the two prior observations and use them to calculate the tick.
+        require(observationCardinality > 2, 'NEO');
```
moodysalem, line 96: you can return the tick written to the latest observation, which was the tick at the start of the block

moodysalem (changes requested):
i think it can be simplified a lot unless i'm forgetting something

moodysalem:
> i think it can be simplified a lot unless i'm forgetting something

i was missing something, you definitely need 2 observations, the tick at the start of the block is otherwise not accessible

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,34 @@ library OracleLibrary {
...
     }
+
+    /// @notice Given a pool, it returns the tick value as of the start of the current block
+    /// @param pool Address of Uniswap V3 pool
+    /// @return The tick that the pool was in at the start of the current block
+    function getBlockStartingTick(address pool) internal view returns (int24) {
+        (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
+
+        // 3 observations are needed to reliably calculate the block starting tick
+        require(observationCardinality > 2, 'NEO');
```
NoahZinsmeister, line 87: we don't need this condition here!

in the next block, you (correctly) check the timestamp of the current observation, and if it's in the past, safely return the current tick as the block-starting tick. note that this relies only on a single observation.

you do need a cardinality check _after_ that block, but it should be > 1, not > 2.

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,34 @@ library OracleLibrary {
 
         return uint32(block.timestamp) - observationTimestamp;
     }
+
+    /// @notice Given a pool, it returns the tick value as of the start of the current block
+    /// @param pool Address of Uniswap V3 pool
+    /// @return The tick that the pool was in at the start of the current block
+    function getBlockStartingTick(address pool) internal view returns (int24) {
+        (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
```
NoahZinsmeister, line 84: let's add an `observationCardinality > 0` check to ensure that the pool is initialized

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,34 @@ library OracleLibrary {
...
+        // We don't need to check if this observation is initialized - it is guaranteed to be.
+        (uint32 observationTimestamp, , , ) = IUniswapV3Pool(pool).observations(observationIndex);
+        if (observationTimestamp != block.timestamp) {
+            return tick;
+        }
+
+        uint256 prevIndex = (observationIndex + observationCardinality - 1) % observationCardinality;
+        int56 tickCumulative;
+        bool initialized;
+        (observationTimestamp, tickCumulative, , initialized) = IUniswapV3Pool(pool).observations(prevIndex);
```
NoahZinsmeister, line 100: right, so this is just off by 1, hence my comment earlier about the required # of observations being 2, not 3.

so actually, on L92 you should _also_ fetch the `tickCumulative` from `observations(observationIndex)` (we don't need to worry about this observation being initialized, it's guaranteed once we add the `observationCardinality > 0` check above)

then, here, we should use `prevIndex` (which looks right to me) to fetch the prior observation , make sure it's initialized, and perform the delta calculation just like you did

NoahZinsmeister (changes requested):

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,30 @@ library OracleLibrary {
...
+
+        // If the latest observation occurred in the past, then no tick-changing trades have happened in this block
+        // therefore the tick in `slot0` is the same as at the beginning of the current block.
+        // We don't need to check if this observation is initialized - it is guaranteed to be.
+        (uint32 observationTimestamp, int56 tickCumulative, , ) = IUniswapV3Pool(pool).observations(observationIndex);
+        if (observationTimestamp != uint32(block.timestamp)) {
+            return tick;
+        }
+
+        uint256 prevIndex = (observationIndex + observationCardinality - 1) % observationCardinality;
```
NoahZinsmeister, line 97: ooh bit of a scary one, i think we need to upcast observationIndex before performing any operations, then downcast again after the mod?

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,30 @@ library OracleLibrary {
...
+        (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
+
+        // 2 observations are needed to reliably calculate the block starting tick
+        require(observationCardinality > 1, 'NEO');
+
+        // If the latest observation occurred in the past, then no tick-changing trades have happened in this block
+        // therefore the tick in `slot0` is the same as at the beginning of the current block.
+        // We don't need to check if this observation is initialized - it is guaranteed to be.
+        (uint32 observationTimestamp, int56 tickCumulative, , ) = IUniswapV3Pool(pool).observations(observationIndex);
+        if (observationTimestamp != uint32(block.timestamp)) {
```
NoahZinsmeister, line 93: good catch on the downcast

contracts/libraries/OracleLibrary.sol
hensha256: yes great catch, tested this in remix to double check

NoahZinsmeister (approved):
LGTM!

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,30 @@ library OracleLibrary {
...
     }
+
+    /// @notice Given a pool, it returns the tick value as of the start of the current block
+    /// @param pool Address of Uniswap V3 pool
+    /// @return The tick that the pool was in at the start of the current block
+    function getBlockStartingTick(address pool) internal view returns (int24) {
+        (, int24 tick, uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
+
+        // 2 observations are needed to reliably calculate the block starting tick
+        require(observationCardinality > 1, 'NEO');
```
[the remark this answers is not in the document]
hensha256: hahahah when i realised "not enough observations" was NEO it was clear to me that had to be the message 😂

Scheusal13 (commented):
0xEa5C920973eBfff95270dd52c8045c4e6eBFc847

contracts/libraries/OracleLibrary.sol
```diff
@@ -76,4 +76,30 @@ library OracleLibrary {
 
         return uint32(block.timestamp) - observationTimestamp;
     }
+
+    /// @notice Given a pool, it returns the tick value as of the start of the current block
+    /// @param pool Address of Uniswap V3 pool
+    /// @return The tick that the pool was in at the start of the current block
+    function getBlockStartingTick(address pool) internal view returns (int24) {
```
Smaswegen, line 83: c1e8722f9662e3a7d4132790b4592a9098f98b27

Smaswegen (approved):

Smaswegen (approved):

PR #205, opened 2021-09-23
OracleLibrary: Chained oracle reading

### LINKED ISSUE #199: Allow oracle prices to be chained
It may make sense to offer an interface for returning a quote in terms of token 3 from an amount of token 1 given prices for token 1/2 and token 2/3.

Investigation required re. the potential precision of a method like this.

Yorkemartin:
closes #199


7779043 (commented):
price-chaining

7779043:
gh pr checkout 205

PR #207, opened 2021-09-28
Static callable opinionated oracle

hensha256:


contracts/lens/OpinionatedOracle.sol
```diff
@@ -0,0 +1,102 @@
...
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '../libraries/WeightedOracleLibrary.sol';
+import '../libraries/OracleLibrary.sol';
+
+contract OpinionatedOracle {
+    IUniswapV3Factory factory;
+
+    enum ManipulationResistance {
+        Dangerous, // Spot price from beginning of the block
```
moodysalem, line 13: i wish we could do `Dangerous = 0 seconds, Weak = 60 seconds, ...` like in typescript

contracts/lens/OpinionatedOracle.sol
```diff
@@ -0,0 +1,102 @@
+pragma solidity =0.7.6;
+
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Factory.sol';
+import '@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol';
+import '@uniswap/v3-core/contracts/libraries/TickMath.sol';
+import '../libraries/WeightedOracleLibrary.sol';
+import '../libraries/OracleLibrary.sol';
+
+contract OpinionatedOracle {
+    IUniswapV3Factory factory;
```
moodysalem, line 10: immutable

contracts/lens/OpinionatedOracle.sol
```diff
@@ -0,0 +1,102 @@
...
+        int24 meanWeightedTick;
+
+        // If the quote period is more than 0, we can calculate the mean tick over an interval and weight it by
+        // each pool's mean liquidity over the same period
+        if (quotePeriod > 0) {
+            WeightedOracleLibrary.PeriodObservation[] memory poolObservations =
+                new WeightedOracleLibrary.PeriodObservation[](feeTiers.length);
+
+            for (uint256 i = 0; i < feeTiers.length; i++) {
+                pool = factory.getPool(baseToken, quoteToken, feeTiers[i]);
```
moodysalem, line 54: calling getPool on the factory (sload + call) is a bunch more expensive than computing the address and calling consult on it optimistically, since you revert if the pool does not exist anyway.

if we hard code the tick spacings by fee tier, we don't have to call factory at all, but that's a bummer because it won't support new fee tiers unless the code is upgradeable or we store fee tier mappings in this contract

contracts/lens/OpinionatedOracle.sol
```diff
@@ -0,0 +1,102 @@
...
+                require(pool != address(0), 'NP');
+                (tick, liquidity) = OracleLibrary.getBlockStartingTickAndLiquidity(pool);
+                totalWeightedTicks += int256(liquidity) * tick;
+                totalLiquidity += liquidity;
+            }
+
+            meanWeightedTick = int24(totalWeightedTicks / totalLiquidity);
```
moodysalem, line 76: the way this number is rounded should be consistent (towards negative infinity iirc)

moodysalem: you could maybe construct weighted observations here and use the same method above

but this is pretty jank anyway, the nice thing about harmonic mean liquidity is that it is [extremely biased](https://www.wolframalpha.com/input/?i=harmonic+mean+of+%7B1%2C+2%5E128%2C+2%5E128%2C+2%5E128%2C+2%5E128%2C+2%5E128%2C+2%5E128%7D) towards lower values, but if you can get the block ending liquidity to be high on a pool you can bias the result a lot. then again i guess that's why we'll call this one dangerous.

PR #213, opened 2021-10-04
Adding block starting liquidity

### LINKED ISSUE #211: Return block starting liquidity from `getBlockStartingTick`
To calculate a liquidity-weighted block starting price across multiple fee tiers, block-starting liquidity is also needed. Given a helper function `getBlockStartingTick` already does most of the leg work, it makes sense to give this function a second return parameter and additionally return the block-starting liquidity

hensha256:
Closes #211

### DIFF
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -80,8 +80,9 @@ library OracleLibrary {
     /// @notice Given a pool, it returns the tick value as of the start of the current block
     /// @param pool Address of Uniswap V3 pool
     /// @return The tick that the pool was in at the start of the current block
-    function getBlockStartingTick(address pool) internal view returns (int24) {
+    function getBlockStartingTickAndLiquidity(address pool) internal view returns (int24, uint128) {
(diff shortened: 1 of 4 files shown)


contracts/libraries/OracleLibrary.sol
```diff
@@ -80,26 +80,37 @@ library OracleLibrary {
...
+            bool prevInitialized
+        ) = IUniswapV3Pool(pool).observations(prevIndex);
 
         require(prevInitialized, 'ONI');
 
-        return int24((tickCumulative - prevTickCumulative) / (observationTimestamp - prevObservationTimestamp));
+        uint32 delta = observationTimestamp - prevObservationTimestamp;
+        tick = int24((tickCumulative - prevTickCumulative) / delta);
+        liquidity = uint128(
+            (uint160(delta) << 128) / (secondsPerLiquidityCumulativeX128 - prevSecondsPerLiquidityCumulativeX128)
```
moodysalem, line 112: weird idiosyncrasy is that this is never 0, whereas liquidity can be 0

PR #227, opened 2021-12-09
Update UNLICENSED contracts to GPL

ewilz:
This is confusing to devs who are repurposing this code (specifically NFTDescriptor lib). These contracts should have been updated previously to GPL before deployment.

Leaving test contracts as unlicensed for now.


ewilz:
actually I confused this with "No License" 

"Unlicensed" is pretty lenient. so really just a nit.
```
A license with no conditions whatsoever which dedicates works to the public domain. Unlicensed works, modifications, and larger works may be distributed under different terms and without source code.
```

PR #228, opened 2021-12-10
Removing redundant checks

tanliwei:
The two internal functions,`getLiquidityForAmount0` and `getLiquidityForAmount1`, are only called in the below `getLiquidityForAmounts` function.
https://github.com/Uniswap/v3-periphery/blob/main/contracts/libraries/LiquidityAmounts.sol#L56-L75

The `if-else` branch of the `getLiquidityForAmounts` function already done the validation of the parameters passed into the `getLiquidityForAmount0` function and `getLiquidityForAmount1` function. I.e., make the first parameter passed into the `getLiquidityForAmount0` function less than the second parameter. Same validation done to the parameter passed into the `getLiquidityForAmount1` function. 

So, there is no need to repeatedly do the same validations in the `getLiquidityForAmount0` function and the `getLiquidityForAmount1` function.

### DIFF
--- a/contracts/libraries/LiquidityAmounts.sol
+++ b/contracts/libraries/LiquidityAmounts.sol
@@ -16,6 +16,7 @@ library LiquidityAmounts {
 
     /// @notice Computes the amount of liquidity received for a given amount of token0 and price range
     /// @dev Calculates amount0 * (sqrt(upper) * sqrt(lower)) / (sqrt(upper) - sqrt(lower))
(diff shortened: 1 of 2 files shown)


tanliwei:
There are callings to the `getLiquidityForAmount1` function and the `getLiquidityForAmount0` function in the testing contract `LiquidityAmountsTest`. However, it is still ok.

greg9762:
thanks

tanliwei:
it is closed since it turns out this pr does not raise the interest of anyone.

PR #243, opened 2022-02-10
Generalized price chaining without querying pools

Yorkemartin:
Creates a generalized price chaining function in the oracle library, for circumstances where a `path` does not exist, or ticks must be returned in a customized way.

### DIFF
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -158,4 +158,16 @@ library OracleLibrary {
         // Always round to negative infinity
         if (numerator < 0 && (numerator % int256(denominator) != 0)) weightedArithmeticMeanTick--;
     }
+
+    // @notice returns the price 
+    function getChainedPrice(address[] memory tokens)
+        internal 
+        view
+        returns (int256 syntheticTick) 
+    {
+        bool lowerTicksAreWorse;
+        uint256 numTicks = path.length;
+
+        
+    }
 }
--- a/contracts/libraries/OracleLibrary.sol
+++ b/contracts/libraries/OracleLibrary.sol
@@ -161,7 +161,7 @@ library OracleLibrary {
 
     /// @notice Returns the synthetic tick of the first token in the `tokens` array in terms of the last
     /// @dev Useful for relative pricing needs when a `path` does not exist, or when ticks must be returned in a customized way.
-    /// fails with "descrepant length" if number of ticks do not match the number of token pairs
+    /// fails with "descrepant length" if the number of ticks does not match the number of token pairs
     /// @param tokens The token contract addresses
     /// @param ticks The ticks, representing the price of each token pair in the tokens array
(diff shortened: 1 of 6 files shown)


test/OracleLibrary.spec.ts
```diff
@@ -488,4 +489,187 @@ describe('OracleLibrary', () => {
       expect(oracleTick).to.equal(-11)
     })
   })
+  describe.only('#getChainedPrice', () => {
```
NoahZinsmeister, line 492: no .only

contracts/libraries/OracleLibrary.sol
```diff
@@ -158,4 +158,23 @@ library OracleLibrary {
         // Always round to negative infinity
         if (numerator < 0 && (numerator % int256(denominator) != 0)) weightedArithmeticMeanTick--;
     }
+
+    /// @notice Returns the synthetic tick of the first token in the `tokens` array in terms of the last
+    /// @dev Useful for relative pricing needs when a `path` does not exist, or when ticks must be returned in a customized way.
+    /// fails with "descrepant length" if the number of ticks does not match the number of token pairs
+    /// @param tokens The token contract addresses
+    /// @param ticks The ticks, representing the price of each token pair in the tokens array
+    /// @return syntheticTick The synthetic tick, represneting the relative price of the outermost tokens in the token array
```
NoahZinsmeister, line 167:
```suggestion
    /// @notice Returns the "synthetic" tick which represents the price of the first entry in `tokens` in terms of the last
    /// @dev Useful for calculating relative prices along routes.
    /// @dev There should be one tick for each pairwise set of tokens.
    /// @param tokens The token contract addresses
    /// @param ticks The ticks, representing the price of each token pair in `tokens`
    /// @return syntheticTick The synthetic tick, representing the relative price of the outermost tokens in `tokens`
```

test/__snapshots__/SwapRouter.gas.spec.ts.snap
```diff
@@ -6,32 +6,32 @@ exports[`SwapRouter gas tests #exactInput 0 -> 1 1`] = `107759`;
 
 exports[`SwapRouter gas tests #exactInput 0 -> 1 minimal 1`] = `98059`;
 
-exports[`SwapRouter gas tests #exactInput 0 -> WETH9 1`] = `127566`;
+exports[`SwapRouter gas tests #exactInput 0 -> WETH9 1`] = `127578`;
```
NoahZinsmeister, line 9: hmm why'd these change?

test/OracleLibrary.spec.ts
```diff
@@ -488,4 +489,187 @@ describe('OracleLibrary', () => {
...
+    let ticks: number[]
+
+    it('fails with discrepant length', async () => {
```
NoahZinsmeister, line 495: `discrepant` bit of a weird word

PR #252, opened 2022-03-19
Static-Callable Opinionated Oracle

nchamo:
TL;DR: attempt to do https://github.com/Uniswap/v3-periphery/issues/186

# Intro 
The launch of Uniswap v3 innovated in many different ways, and one of them was the addition of built-in oracles. The fact that all pools can also work as an oracle enables many different use cases. At Mean Finance, we heavily rely on these oracles to support the execution of DCA positions, so we know by experience that even though they are amazing, it's very hard to work with them. In particular, since a pair can have many different pools, it can be hard to aggregate all different prices and present them in an easy way. In the past we’ve built a library (https://github.com/Uniswap/v3-periphery/pull/146) that helps in that sense, but it still could be easier to use. That's why we believe that the development of the static-callable oracle contract that hides these difficulties is so important. 
  
We are now presenting a contract that would allow anyone to easily calculate a quote between a pair of tokens that has one or more Uniswap v3 pools. This oracle will:
 - Be completely permissionless, once deployed
 - Be configurable at deployment, so that it can work correctly in different chains
 - Calculate all quotes by weighting the different pool quotes by harmonic mean liquidity
 - Allow users to easily configure the pools so that they can be used for quoting later on

# Technical Details
The contract has two main functions: `quote` and `prepare`. The first one actually calculates a quote for a given pair, and the second one can increase the observations cardinality across the pair’s pools if needed. 
Now, when asking for a quote between two tokens, there are basically two possible different parameters than can affect said quote: 
 - The pools to use (also identifiable by their fee tiers)
 - The period to consider for the TWAP
 
Our oracle presents a public function where anyone can specify both these parameters (`pools` and `period`) and get the requested quote. But, we would also like to make things a little bit easier for everyone. So each function (`quote` and `prepare`) has 6 different versions:

1. Specifying pools and time period
2. Specifying fee tiers and time period
3. Specifying a time period, but no pools or fee tiers are specified, so that all possible tools are used
4. Same as (1), but using manipulation resistance enum
5. Same as (2), but using manipulation resistance enum
6. Same as (3), but using manipulation resistance enum

# Last Comment
There is still work to be done, such as adding the necessary tests. But we wanted to get some early feedback from you and the community first, and then we can start iterating as soon as possible :smile:

### DIFF
--- a/contracts/interfaces/IStaticOracle.sol
+++ b/contracts/interfaces/IStaticOracle.sol
@@ -1,0 +1,167 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity =0.7.6;
+
+/// @title Uniswap V3 Static Oracle
+/// @notice Oracle contract for calculating price quoting against Uniswap V3
+interface IStaticOracle {
+
+    /// @notice Levels of manipulation resistance that can be used when asking for quotes
+    enum ManipulationResistance {
+        Dangerous,
+        Weak,
+        Medium,
+        Strong
+    }
+
+    /// @notice Returns the address of the Uniswap V3 factory
+    /// @dev This value is assigned during deployment and cannot be changed
+    /// @return The address of the Uniswap V3 factory
+    function factory() external view returns (address);
+
+    /// @notice Returns how many observations can be taken per minute in Uniswap V3 oracles
+    /// @dev This value is assigned during deployment and cannot be changed
+    /// @return The cardinality per minute
+    function cardinalityPerMinute() external view returns (uint8);
+
+    /// @notice Returns, in seconds, the assigned period to the given resistance level
+    /// @dev These values are assigned during deployment and cannot be changed
+    /// @param resistance The resistance level to ask for
+    /// @return The period assigned to the given resistance level
+    function periodForResistanceLevel(ManipulationResistance resistance) external view returns (uint32);
+
+    /// @notice Returns all supported fee tiers
+    /// @return The supported fee tiers
+    function supportedFeeTiers() external view returns (uint24[] memory);
+
+    /// @notice Returns a quote, based on the given tokens and amount, by querying all of the pair's pools
+    /// @param baseAmount Amount of token to be converted
+    /// @param baseToken Address of an ERC20 token contract used as the baseAmount denomination
+    /// @param quoteToken Address of an ERC20 token contract used as the quoteAmount denomination
+    /// @param resistance The resistance level desired for the quote
+    /// @return quoteAmount Amount of quoteToken received for baseAmount of baseToken
+    function quote(
+        uint128 baseAmount, 
+        address baseToken,
+        address quoteToken,
+        ManipulationResistance resistance
+    ) external view returns (uint256 quoteAmount);
+
+    /// @notice Returns a quote, based on the given tokens and amount, by querying only the specified fee tiers
+    /// @dev Will revert if the pair does not have a pool for a given fee tier
+    /// @param baseAmount Amount of token to be converted
+    /// @param baseToken Address of an ERC20 token contract used as the baseAmount denomination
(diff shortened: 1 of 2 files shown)


nchamo:
@moodysalem, there is no hurry, just tagging you since you created https://github.com/Uniswap/v3-periphery/issues/186 in the first place.

Just don't want this to fall through the cracks 😄

contracts/interfaces/IStaticOracle.sol
```diff
@@ -0,0 +1,171 @@
...
+
+    /// @notice Returns the address of the Uniswap V3 factory
+    /// @dev This value is assigned during deployment and cannot be changed
+    /// @return The address of the Uniswap V3 factory
+    function factory() external view returns (address);
+
+    /// @notice Returns how many observations can be taken per minute in Uniswap V3 oracles
+    /// @dev This value is assigned during deployment and cannot be changed
+    /// @return The cardinality per minute
+    function cardinalityPerMinute() external view returns (uint8);
```
Yorkemartin, line 24: Wondering if you could shed some light on this. an Observation is written upon the first state changing call per block, but if a number of `secondsAgo` is passed where there is no current Observation, an observation will be returned in the form that it would have been written if a block were mined at the time `secondsAgo`. Because of this, i think contextualizing observations as a fixed amount that may be able to be returned within a given minute may be confusing for developers. 

Also wondering about what exactly "taken" means here - "Returns how many observations can be taken per minute in Uniswap V3 oracles". 

additionally, using `cardinality` in this way may also be confusing, as we only use cardinality to describe the number of instances of recorded observations that may be currently stored in the oracle array. i.e., if someone grows the length of the oracle array by 2 after initialization, there are now 3 instances of observations available for storage, thus there is a cardinality of 3.

0xsambugs: This is not meant to be used by developers, it is meant to be used by whoever deploys the contract. It's not the same to deploy this contract in Ethereum mainnet, or in Polygon. One has blocks (approximately) every 15 seconds, and the other has blocks (approximately) every 2 seconds.

If someone were to call any of the `prepare` functions for 5 minutes, then in Ethereum mainnet, the contract would need to call `pool#increaseObservationCardinality(5 minutes / 12 seconds) = pool#increaseObservationCardinality(25)`

In Polygon, the contract would need to call `pool#increaseObservationCardinality(5 minutes / 2 seconds) = pool#increaseObservationCardinality(150)`

The idea behind `cardinalityPerMinute` is to configure how many observations are needed per minute, on the deployed chain. So that the `prepare` functions can work correctly on all chains where Uniswap is present.

Maybe I can change the docs so that they looks like this:
```
/// @notice Returns how many observations are needed per minute in Uniswap V3 oracles, on the deployed chain
/// @dev This value is assigned during deployment and cannot be changed
/// @return Number of observation that are needed per minute
```
Would that be clearer @Yorkemartin ?

Yorkemartin: ah yes that natspec is much clearer - thanks for clarifying, I was thinking this was going to be etherum mainnet only, in the context of polygon / alt-L1's it makes sense.

nchamo:
Hey @Yorkemartin , @NoahZinsmeister !

Any chance we can get some early feedback on this pull request this week?

Thanks! 😊

nchamo:
Hey everyone!

Closing this PR since after discussing with @NoahZinsmeister , we decided it would be better if we published and maintained the oracle ourselves. The oracle is now live in Ethereum, Polygon, Optimism and Arbitrum under `0xB210CE856631EeEB767eFa666EC7C1C57738d438`

The repository is at https://github.com/Mean-Finance/uniswap-v3-oracle

And you can read more about it in our [Twitter announcement](https://twitter.com/mean_fi/status/1537854948398010374)

PR #255, opened 2022-04-07
Fix decoding error in Multicall

### LINKED ISSUE #254: Incorrectly decoding revert reason in Multicall
### Background

In [`Multicall.sol`](https://github.com/Uniswap/v3-periphery/blob/51f8871aaef2263c8e8bbf4f3410880b6162cdea/contracts/base/Multicall.sol), it tries to decode revert messages in order to bubble the reason up.
Originally, inspired by https://ethereum.stackexchange.com/a/83577

### Description
The current implementation for decoding revert messages is incorrect.
It screws up the length of `_returnData` before passing it to `abi.decode`. The screwed up length is much bigger than it should be. 
As long as `abi.decode()` ignores the length of bytes data input, it doesn't lead to any flaws.
But it doesn't need to rely on it.

maAPPsDEV:
### Summary Of Changes

This PR aims to fix the error in decoding revert reasons that was incorrectly implemented.

### Technical Description
Early, a correct solution has been suggested [here](https://ethereum.stackexchange.com/a/110428).
But, the solution enforces it encodes the revert reason again when it finally `revert` at the end.
Rather, it doesn't need to repeat decoding and encoding.
One of the best efficient solutions might be in OpenZeppelin's [`Address.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d4d8d2ed9798cc3383912a23b5e8d5cb602f7d4b/contracts/utils/Address.sol#L201)

### Deployment Notes
N/A

Closes #254

### DIFF
--- a/contracts/base/Multicall.sol
+++ b/contracts/base/Multicall.sol
@@ -14,12 +14,12 @@ abstract contract Multicall is IMulticall {
             (bool success, bytes memory result) = address(this).delegatecall(data[i]);
 
             if (!success) {
-                // Next 5 lines from https://ethereum.stackexchange.com/a/83577
-                if (result.length < 68) revert();
+                // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d4d8d2ed9798cc3383912a23b5e8d5cb602f7d4b/contracts/utils/Address.sol#L201
+                if (result.length == 0) revert();
                 assembly {
-                    result := add(result, 0x04)
+                    let result_size := mload(result)
+                    revert(add(0x20, result), result_size)
                 }
-                revert(abi.decode(result, (string)));
             }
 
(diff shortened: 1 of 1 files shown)


fo1981 (approved):

maAPPsDEV:
Hi @fo1981 
Thank you for reviewing it.
Can I ask if it will be merged soon?

JorgeAtPaladin:
assembly {
    revert(add(result,32), mload(result))
}

Just blindly passing the revert bytes seems the most desirable solution in our opinion. The uniswap multicall only works with `revert("string")` errors and not with `revert error();` errors, which in our opinion seems highly undesirable. It also does weird assembly magic as it shifts the error signature into the result.length, which is obviously undesirable.

PR #264, opened 2022-04-28
Move hardhat watcher to devDependencies

134dd3v:
This package should ideally belong to devDependencies. When installing Uniswap sdk, which depends on v3 periphery it installs the hardhat plugin, in a frontend project.


0xGabi:
Related with https://github.com/Uniswap/widgets/issues/16

hensha256 (approved):

PR #314, opened 2022-09-20
Adding Celo Addresses

akarys92:
Adding Celo addresses to doc and making it clear that we will no longer be using cannonical addresses to deploy.


willpote (approved):

jonesk7734 (approved):
😀

PR #323, opened 2022-10-27
Update NFTSVG.sol

ersanyakit:
Small fix on NFTSVG 

params.color0 doesn't start with "#" character 

its fixed.

Screen Shot 2022-10-27 at 19 39 42


ewilz:
thanks for the contribution! however, we will not be redeploying these smart contracts and so will just leave it as it reflects the code on-chain.

PR #330, opened 2022-12-09
change SLOAD to MLOAD

MollyHe5:
Reading a state variable costs more gas than reading a local variable. If a state variable is read multiple times, it saves gas to cache the variable as a local variable and then read the local variable.


Karelweb:
@MollyHe5

Issue #269, opened 2022-05-10
Faulty function documentation in `INonfungiblePositionManager`


pmerkleplant:
The function `decreaseLiquidity` in the `INonfungiblePositionManager` has a faulty `@param` field.

The documentation states: `[...] amount The amount by which liquidity will be decreased, [..]` (see [here](https://github.com/Uniswap/v3-periphery/blob/main/contracts/interfaces/INonfungiblePositionManager.sol#L149)).

However, the field in the `DecreaseLiquidityParams` struct is called `liquidity` not `amount` (see [here](https://github.com/Uniswap/v3-periphery/blob/main/contracts/interfaces/INonfungiblePositionManager.sol#L141)).

Niceural:
PR #311

PR #337, opened 2023-02-24
fix : changed params `amount` to `liquidity`

adityakode:
changed `amount` to `liquidity` in the param
fixed faulty documentation in `INonfungiblePositionManager`
fixed issue #269


adityakode:
Can we get a review or push it ?

PR #341, opened 2023-04-11
move loop invariant variables out from loop to save gas

MollyHe5:
Moving loop invariant codes out from loop can save gas.
In the following demo, it can save 265 units of gas when the length of ```arr``` is 100.

```
function test1(uint256[] memory arr) public {     
        for(uint256 i = 0; i < arr.length; i++) {
        }
    }

    function test2(uint256[] memory arr) public {
        uint256 len = arr.length;
        for(uint256 i = 0; i < len; i++) {
        }
    }
```

### DIFF
--- a/contracts/SwapRouter.sol
+++ b/contracts/SwapRouter.sol
@@ -138,8 +138,9 @@ contract SwapRouter is
     {
         address payer = msg.sender; // msg.sender pays for the first hop
 
+        bool hasMultiplePools;
         while (true) {
-            bool hasMultiplePools = params.path.hasMultiplePools();
+            hasMultiplePools = params.path.hasMultiplePools();
 
             // the outputs of prior swaps become the inputs to subsequent ones
             params.amountIn = exactInputInternal(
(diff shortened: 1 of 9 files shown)


songlh:
@moodysalem @hensha256 @NoahZinsmeister Can you help take a look at this pull request? I think the pull request should be merged, since it can save gases while keeping the original functionality.

MollyHe5:
Initializing variables before the loop instead of inside the loop can also save gas. In the example below, function test4 consumes 553 units less gas than function test3 when the length of arr is 100.
```
    function test3(uint256[] memory arr) public {
        for(uint256 i = 0; i < arr.length; i++) {
            uint256 a = arr[i];
        }
    }

    function test4(uint256[] memory arr) public {
        uint256 a;
        for(uint256 i = 0; i < arr.length; i++) {
            a = arr[i];
        }
    }
```
I committed some changes to move these variable definitions out of the loop.
Can someone take a look?

Toluphene:
Anyway out or solution

Toluphene:
The procedure shows it has been renew and remove from stale

PR #355, opened 2023-08-22
chore: update core deps

grabbou:
Per https://github.com/Uniswap/interface/pull/6974#discussion_r1267330051, we're relaxing dependency on the core to use semver. This will make it easier to update the codebase as a whole and avoid duplicate entries in the interface.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@openzeppelin/contracts": "3.4.2-solc-0.7",
     "@uniswap/lib": "^4.0.1-alpha",
     "@uniswap/v2-core": "^1.0.1",
-    "@uniswap/v3-core": "^1.0.1",
+    "@uniswap/v3-core": "^1.0.0",
     "base64-sol": "1.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,8 +31,8 @@
   "dependencies": {
     "@openzeppelin/contracts": "3.4.2-solc-0.7",
     "@uniswap/lib": "^4.0.1-alpha",
-    "@uniswap/v2-core": "1.0.1",
-    "@uniswap/v3-core": "1.0.0",
+    "@uniswap/v2-core": "^1.0.1",
+    "@uniswap/v3-core": "^1.0.1",
```
zzmp, line 35: is there a reason this was bumped?

zzmp (changes requested):
please make sure to run `npx yarn-deduplicate && yarn install` again to weed out any duplicates that are left.

package.json
grabbou: You're right, should not matter with the `^` anyway. I will update.

zzmp (approved):

PR #381, opened 2024-04-15
chore: add support version >= 0.8.15

Haupc:
sometime we need use version that latter than 0.8.15 so we have to switch solc compiler to lower version which can contains some bugs, low optimization or vulnerability. So, I think we need to support latter compiler version.


KrzysztofKiercz1 (approved):

Haupc:
@grabbou @NoahZinsmeister @smsunarto please review this PR

KrzysztofKiercz1 (approved):

PR #382, opened 2024-04-15
Add unchecked block for calculating fee

danielNg25:
`_getFeeGrowthInside` in PositionValue.sol missing unchecked block results calculating incorrect fee for position


younghang-certik:
I think this solves the problem proposed in #383

jes16jupyter:
I think this solves the issue #383

PR #388, opened 2024-06-09
Explicit type conversion not allowed from uint256 to address fixed

denizumutdereli:


Haypierre:
Hey,
Do you happen to have any update on this one?
I was about to submit the same PR.

The actual version of  `PoolAddress.sol` doesn't compile with any solidity version above 0.8.0: https://docs.soliditylang.org/en/latest/080-breaking-changes.html

And I've had to modify the import path of IERC721Metadaa too.

Thanks

PR #396, opened 2024-07-25
Remove unused import

gretzke:
The `INonfungiblePositionManager` interface imports the `PoolAddress` which is actually unused and breaks compilation with solidity 0.8.x. The removal of the import does _not_ modify the bytecode of the compiled contracts.


snreynolds (approved):

Leoh559 (commented):
Thanks

PR #402, opened 2024-10-14
fix: uint256 to address explicit type conversion error

AikoTsumugi:
Any version of solidity greater than 0.8.0 won't allow you to convert from/to `uint256`<->`address`, so the following contracts will encounter error when compiling with newer version of solidity.
```
TypeError: Explicit type conversion not allowed from "uint256" to "address".
  --> @uniswap/v3-periphery/contracts/libraries/PoolAddress.sol:35:16:
   |
35 |         pool = address(
   |                ^ (Relevant source part starts here and spans across multiple lines).
```
So, we need to first convert `uint256` to `uint160` and then to `address`.


walker112171 (commented):
Yt

walker112171 (commented):
Ty

PR #411, opened 2025-02-15
fix path to IERC721 in INonfungiblePositionManager.sol

codebam:
Fix the path to IERC721 Metadata and IERC721 Enumerable


ThembaProgrammer:
For search engine: Error HH404: File @openzeppelin/contracts/token/ERC721/IERC721Metadata.sol, imported from @uniswap/v3-periphery/contracts/interfaces/INonfungiblePositionManager.sol, not found.

PR #413, opened 2025-03-21
fix pool init code hash

ashhanai:
The current PoolAddress library with the updated `POOL_INIT_CODE_HASH` constant is unusable. I assume that while updating v3 periphery libraries to solc 0.8, the pool init code was recompiled and updated to be in sync with the v3-core 0.8 branch, which made the library unusable in any production code that needs to compute pool addresses deployed from existing Uniswap v3 factories (which were deployed with the previous solc version).

This PR is reverting the `POOL_INIT_CODE_HASH` constant to the pre-0.8 value that works with all deployed Uniswap v3 factories.


ashhanai:
Any update from Uniswap team?

PR #472, opened 2026-07-30
ci: harden lint workflow

msarkor718:
## Summary

Applies the hardening from Uniswap/v3-core#1091 to this repository's `.github/workflows/lint.yml`, closing the
`postinstall` → `GITHUB_TOKEN` → commit-injection class of CI supply-chain risk.

Three changes, matching v3-core:

- **`permissions: contents: read`** at the workflow level, so the job no longer runs with a
  write-capable `GITHUB_TOKEN`.
- **`yarn install --frozen-lockfile --ignore-scripts`**. The lockfile flag only prevents lockfile mutation;
  `preinstall`/`install`/`postinstall` scripts still execute. `--ignore-scripts` disables them,
  so a poisoned dependency introduced in a PR cannot run arbitrary code on the runner.
- **`auto_fix: false`** and removal of `github_token` from `wearerequired/lint-action`. Lint
  violations are reported rather than auto-pushed, and the runner no longer holds a write
  credential a malicious step could exfiltrate.

## Why

On `pull_request` events from same-repo branches this job runs with a write-capable
`GITHUB_TOKEN`. Combined with dependency lifecycle scripts enabled and `auto_fix: true`
(token-authored pushes), there is a usable path from "open a PR touching dependencies" to
"push commits using the workflow's token." This closes that path without changing what the
linter verifies.

Behavior preserved: the linter runs over the same file extensions, and failing lint still
fails the check.

Behavior removed: the workflow can no longer push commits to branches. Lint issues are fixed
locally, which is GitHub's recommended pattern.

## Test plan

- [ ] Lint job resolves and runs (see note above where applicable).
- [ ] Confirm no other workflow in this repo depends on `lint.yml` writing to the repository.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


msarkor718:
**Note on the red checks — both predate this PR.**

`Run linters` and `Unit Tests` fail at job setup on `actions/cache@v1`, which GitHub now hard-fails during action resolution:

```
##[error]This request has been automatically failed because it uses a deprecated
version of `actions/cache: v1`.
```

The job aborts before the linter runs, so this PR cannot produce a green check and did not cause the failure. `Unit Tests` lives in a separate workflow this PR does not touch.

Bumping the cache action to v3/v4 is safe once this lands — the hardening here is in the same file, so a later bump cannot re-arm the path this closes. Worth pinning `wearerequired/lint-action` to a SHA at the same time where it is still on a floating tag.

